PR #24, opened 2017-11-23
Move MultiFileReader to go-ipfs-cmdkit

lgierth:
The other end of this has already happened in ipfs/go-ipfs-cmdkit#3


keks:
Damn, I totally missed this one. LGTM - @lgierth this is supposed to be merged, right?

lgierth:
It's not a problem, it's dead code

PR #34, opened 2017-12-04
Fix error and cancelation handling in legacy

keks:
Properly handle errors and cancelation.

I must have missed that I can do something meaningful with the error here. Doing that now.

### DIFF
--- a/command.go
+++ b/command.go
@@ -70,9 +70,14 @@ var ErrIncorrectType = errors.New("The command returned a value with a different
 // Call invokes the command for the given Request
 func (c *Command) Call(req Request, re ResponseEmitter) (err error) {
 	// we need the named return parameter so we can change the value from defer()
-	defer re.Close()
+	defer func() {
+		if err == nil {
+			re.Close()
+		}
+	}()
 
-	cmd, err := c.Get(req.Path())
+	var cmd *Command
+	cmd, err = c.Get(req.Path())
 	if err != nil {
 		return err
 	}
--- a/legacy.go
+++ b/legacy.go
@@ -63,8 +63,7 @@ func (rw *responseWrapper) Output() interface{} {
 
 					if err == io.EOF || err == context.Canceled {
 						return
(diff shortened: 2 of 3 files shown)


keks:
This will fix https://github.com/ipfs/go-ipfs/issues/4360

kevina:
I will want to test this on go-ipfs, is there a gx-published hash available?  If not I will do it myself.

kevina:
I was able to just use `gx-go rewrite --undo` to test this.

In ipfs/go-ipfs#4360 I am now getting:

```
$ ipfs repo verify
^C1 blocks processed.
Error: context canceled
$ ipfs repo verify
^C3 blocks processed.
Error: expected type <-chan interface {}, got <nil>
$ ipfs repo verify
^C5 blocks processed.
Error: context canceled
$ echo $?
1
$ ipfs repo verify
^C1 blocks processed.
Error: context canceled
$ ipfs repo verify
^C2 blocks processed.
Error: context canceled
$ ipfs repo verify
^C9 blocks processed.
Error: context canceled
$ ipfs repo verify
^C1 blocks processed.
Error: context canceled
$ ipfs repo verify
^C blocks processed.
Error: expected type <-chan interface {}, got <nil>
$ echo $?
1
```
So I still get the `Error: expected type <-chan interface {}, got <nil>` but at least now the exit code is always non-zero.

keks:
Looks like I didn't fix the race, only made the correct result win it more often. Looking into it again.

Also regarding publishing: I don't want to pin development versions because that just accumulates a lot of soon-obsolete data and cleaning that up is too much maintainence overhead. However, TIL that instead you can use `gx-go link <current cmdshash>` and gx will use symlinks to make it use the dvcspath repo instead of the fixed gx'd version. You might need to upgrade gx and gx-go though.

kevina:
Okay I am now always getting `Error: context canceled` and the code looks good as far as I can tell.

PR #39, opened 2017-12-07
Create Opentracing Spans for HTTP Requests

frrist:
A tracer provides an interface for Span creation and SpanContext
propagation.
A Span represents an active, un-finished event in the OpenTracing
system.
The tracer is used throughout the context of commands request. A Root-Span is created to represent each ServeHTTP event and is added to the context of the request. Subsequent Child-Spans create from this context will be appended to the Root-Span and upon completion will represent a "trace" of a ServeHTTP event.

License: MIT
Signed-off-by:Forrest Weston <Forrest@protocol.ai>

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -132,7 +132,12 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	defer cancel()
 
-	req.Context = logging.ContextWithLoggable(req.Context, loggables.Uuid("requestId"))
+	req.Context = log.EventBeginInContext(req.Context, "ServeHTTP",
+		loggables.Uuid("requestId"), logging.LoggableMap{
+			"httpUrl":    r.URL.String(),
+			"httpMethod": r.Method,
+		})
+
 	if cn, ok := w.(http.CloseNotifier); ok {
 		clientGone := cn.CloseNotify()
 		go func() {
@@ -140,6 +145,7 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			case <-clientGone:
 			case <-req.Context.Done():
 			}
+			logging.MaybeFinishEvent(req.Context)
 			cancel()
 		}()
 	}


http/handler.go
```diff
@@ -117,6 +122,25 @@ func (i Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	i.corsHandler.ServeHTTP(w, r)
 }
 
+// Init returns an instance of Jaeger Tracer that samples 100% of traces and logs all spans to stdout.
+func InitGlobalTracer(service string) (opentracing.Tracer, io.Closer) {
```
ForrestWeston: Need to find a more appropriate location for this method, possibly add it to go-log?

ForrestWeston:
[Jaeger Tracer Docs](https://jaeger.readthedocs.io/en/latest/)
To start a docker image of the tracer
```
docker run -p5775:5775/udp -p6831:6831/udp -p6832:6832/udp \                                                                                                                                                                                       Thu Dec  7 19:54:38 UTC 2017
    -p5778:5778 -p16686:16686 -p14268:14268 -p9411:9411 jaegertracing/all-in-one:0.8.0
```
Then navigate to `http://localhost:16686` to access the Jaeger UI.

http/handler.go
```diff
@@ -164,15 +179,21 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
 	d := re.(Doner)
 	select {
 	case <-d.Done():
 	case <-req.Context.Done():
-		log.Errorf("waiting for http.Responseemitter to close but then %s", req.Context.Err())
+		errMsg := "waiting for http.Responseemitter to close but then " + req.Context.Err().Error()
+		log.Errorf("%s", errMsg)
```
magik6k: Could probably just do `log.Error(errMsg)`

http/handler.go
```diff
@@ -164,15 +179,21 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		rootSpan.LogKV("error", err.Error())
 		return
 	}
 
 	d := re.(Doner)
 	select {
 	case <-d.Done():
 	case <-req.Context.Done():
-		log.Errorf("waiting for http.Responseemitter to close but then %s", req.Context.Err())
+		errMsg := "waiting for http.Responseemitter to close but then " + req.Context.Err().Error()
```
magik6k: Using `fmt.Sprintf` with `req.Context.Err()` would be a bit safer, though I don't think that `req.Context.Err()` should ever return nil.

http/handler.go
```diff
@@ -133,7 +136,19 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	req.Context = logging.ContextWithLoggable(req.Context, loggables.Uuid("requestId"))
+	reqId := loggables.Uuid("requestId")
+	req.Context = logging.ContextWithLoggable(req.Context, reqId)
+
+	// Create rootSpan for incoming request and tag
+	// All spans created after this point will be children of rootSpan
+	rootSpan := opentrace.StartSpan("ServeHTTP")
+	defer rootSpan.Finish()
+	ctx = opentrace.ContextWithSpan(ctx, rootSpan)
```
Stebalien: I don't think we're actually doing anything with this context. I believe we should be replacing `req.Context`.

http/handler.go
```diff
@@ -133,7 +136,19 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	defer cancel()
 
-	req.Context = logging.ContextWithLoggable(req.Context, loggables.Uuid("requestId"))
+	reqId := loggables.Uuid("requestId")
+	req.Context = logging.ContextWithLoggable(req.Context, reqId)
```
Stebalien: Opening two logging events feels a bit redundant. Does opentrace not replace logging?

frrist: This doesn't open logging events, it is creating a loggable requestId and associating it with the context so that events logged with the context will contain the requestId loggable. We then use the requestId below as a tag on the `rootSpan`. This is a bit clunky though..

Stebalien: Events may not be the right word but we're still maintaining two independent logging "contexts". This is not a viable long-term solution and feels like a hack.

Is there any reason we're not using `logging.EventBeginInContext`? Are we not exposing something we need from that? IIRC, you changed that to open opentracing spans.

frrist:
I'm going to close this until the detail in https://github.com/ipfs/go-log/pull/20 are worked out completely. Will create a new branch when ready.

Issue #35, opened 2017-12-05
Add minimial Makefile, use standard test script


kevina:
For libp2p projects that don't have a Makefile I have been using this minimal one:
```
gx:
	go get github.com/whyrusleeping/gx
	go get github.com/whyrusleeping/gx-go

deps: gx
	gx --verbose install --global
	gx-go rewrite

publish:
	gx-go rewrite --undo

```

I have also converted them to use this standard test script: https://github.com/ipfs/ci-helpers/blob/master/travis-ci/run-standard-tests.sh

Should I create a p.r. for this?

Stebalien:
SGTM.

keks:
Yup sound good!

kevina:
This is already done in #49.  The only thing I noticed was that the change added the `-race` to the tests and wonder if it was intentional or not.

Stebalien:
Probably. It's good to have anyways.

Thanks for noticing this.

Issue #23, opened 2017-11-22
Fix CI


keks:
Currently the tests don't run because we depend on go-ipfs and that depends on a lot of gx stuff.

Depends on #20.

Note: To run the tests you have to gxify the whole thing:

```
gx-go rewrite
go test ./...
gx-go rewrite --undo
```

keks:
@Stebalien can I close this? Travis seems to be happy.

Stebalien:
Ah, yeah, this was that circular dependency thing.

PR #44, opened 2017-12-13
fix deadlock on error

### LINKED ISSUE #43: Deadlock when command returns an error
Reproducible by running t0110-gateway.sh. Test 31 fails ("refs IPFS directory file through readonly API succeeds").

See https://github.com/ipfs/go-ipfs/pull/4453

Cause: Command fails and then we wait for it to succeed.

Stebalien:
fixes #43


whyrusleeping:
Why doesnt CI pass?

keks:
@whyrusleeping kevina said he'd work on that https://github.com/ipfs/go-ipfs-cmds/issues/35

Stebalien:
@keks that's unrelated. CI doesn't pass because of the circular dependency: #23

PR #49, opened 2017-12-21
Blocked - Changes for cmds 1.0

keks:
These are the changes for go-ipfs-cmds 1.0.

The aim is to decouple this from go-ipfs, so it can be used for other projects, as well as making it easier to use. Changes include:
- don't depend on go-ipfs/commands
  - moving the legacy layer to go-ipfs/commands/legacy
  - remove OldSubcommands from `Command` struct
- removing the Context struct and replacing it with the `interface{}` variable `env`.
  - Instead of attaching it to the Request it is passed to PreRun and Run directly.
  - env is checked for special behaviour:
    - if it has a method `LogRequest(*Request) func()` it calls it when a request comes in and calls the function returned by it when the request is over. go-ipfs now uses the Context in go-ipfs/commands and has added an according method to interface the request log.
    - if it has a method `Close()`, that method will be called when the request is over.
    - if it has a method `RootContext() context.Context`, it will be called to get a root context that the request context will be derived from.
- removing some dead code
- Request is now a struct
  - Before it was an interface that had a lot of getters and setters. There was no point in doing it that way, so I moved it to being a struct.
  - That's why we now pass a `*Request` instead of a `Request` and accessing fields drop the `()` everywhere. A lot of diff noise, I know.
- Add the Executor interface and two implementations (local and http). Executing a request using http will cal `PreRun` and `PostRun` and send it to the ipfs daemon, executing it locally will call the functions `PreRun`, `PostRun` and `Run`.
- don't use go-ipfs' godep'd version of levenshtein but the gx'd version
- update to go-ipfs-cmdkit 1.0
	- move predefined options from cmdkit here
- don't implicitly use predefined options on everything, instead make the user add them to their root command
- http: make API prefix (/api/v0) configurable

Blocked by [cmdkit 1.0](https://github.com/ipfs/go-ipfs-cmdkit/pull/9). Once that landed the tests should pass (CI doesn't use gx at the moment so it doesn't use cmdkit 1.0)


chan.go
```diff
@@ -99,7 +99,7 @@ func (r *chanResponse) Next() (interface{}, error) {
 		}
 	case <-ctx.Done():
 		close(r.done)
-		return nil, r.req.Context().Err()
+		return nil, r.req.Context.Err()
```
whyrusleeping: Shouldnt this return `ctx.Err()` now?

chan.go
```diff
@@ -125,7 +125,7 @@ func (r *chanResponse) RawNext() (interface{}, error) {
 		return nil, io.EOF
 	case <-ctx.Done():
 		close(r.done)
-		return nil, r.req.Context().Err()
+		return nil, r.req.Context.Err()
```
whyrusleeping: same, ctx.Err()

cli/helptext.go
```diff
@@ -225,9 +246,9 @@ func ShortHelp(rootName string, root *cmds.Command, path []string, out io.Writer
 func generateSynopsis(cmd *cmds.Command, path string) string {
 	res := path
 	for _, opt := range cmd.Options {
-		valopt, ok := cmd.Helptext.SynopsisOptionsValues[opt.Names()[0]]
+		valopt, ok := cmd.Helptext.SynopsisOptionsValues[opt.Name()]
 		if !ok {
-			valopt = opt.Names()[0]
+			valopt = opt.Name()
```
whyrusleeping: We only have one name for options now?

keks: We have one main name and `n` aliases. That way you always use the long form when accessing the option in your code, but you can set the option using any of the provided names.

cli/parse.go
```diff
@@ -14,265 +15,182 @@ import (
...
+		if (req.Path[1] == "add" && req.Options["default"] == true) ||
+			(req.Path[1] == "rm" && req.Options["all"] == true) {
 			stdin = nil
 		}
 	}
 
-	stringArgs, fileArgs, err := ParseArgs(req, stringVals, stdin, cmd.Arguments, root)
-	if err != nil {
-		return req, cmd, path, err
+	if err = parseArgs(req, root, stdin); err != nil {
```
whyrusleeping: style thing, but if i'm using this pattern, I always make sure to use `if err := ` so we don't accidentally reuse the outer scoped variable

cli/helptext.go
keks: That made the cmds code simpler and the actual commands easier to read because you have a consistent name for every option.

cli/parse.go
keks: What's wrong with reusing that variable? All we do is set it right before we return.

chan.go
keks: 👍

chan.go
keks: 👍

cli/parse.go
```diff
@@ -341,29 +265,142 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmdkit.Argument, recur
...
 			if argDef.Required {
-				return nil, nil, fmt.Errorf("Argument '%s' is required", argDef.Name)
+				return fmt.Errorf("argument %q is required", argDef.Name)
+			}
+		}
+	}
+
+	req.Arguments = stringArgs
+	if len(fileArgs) > 0 {
+		req.Files = files.NewSliceFile("", "", filesMapToSortedArr(fileArgs))
```
whyrusleeping: map to sorted array?

cli/parse.go
```diff
@@ -341,29 +265,142 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmdkit.Argument, recur
...
+
+func parseOpt(opt, value string, opts map[string]cmdkit.Option) interface{} {
+	optDef, ok := opts[opt]
+	if !ok {
+		panic(fmt.Errorf("unknown option %q", opt))
+	}
+
+	v, err := optDef.Parse(value)
+	if err != nil {
+		panic(err)
```
whyrusleeping: panics?

keks: yup, that's common in recursive descent parsers. I catch it in the root.

whyrusleeping: If there are okay (program invariants or something) lets make sure to document why

cli/parse.go
```diff
@@ -341,29 +265,142 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmdkit.Argument, recur
...
+
+type kv struct {
+	Key   string
+	Value interface{}
+}
+
+func parseShortOpts(cmdline []string, i int, optDefs map[string]cmdkit.Option) ([]kv, int) {
+	k, vStr, ok := splitkv(cmdline[i][1:])
+	var (
+		j   int
```
whyrusleeping: why declare j here? could just do it down in the for loop, no?

cli/parse.go
keks: I'm pretty sure I didn't come up with this and took it from the existing code. Probably from [here](https://github.com/ipfs/go-ipfs-cmds/pull/49/files/e58b7d68f5b9fa926fbe82eb9d037be7e5c34e51#diff-3be2b56996dc8a3b09575e16f3cc9ef0L364)

cli/parse.go
```diff
@@ -341,29 +265,142 @@ func parseArgs(inputs []string, stdin *os.File, argDefs []cmdkit.Argument, recur
...
+		kvs = append(kvs, kv{
+			Key:   k,
+			Value: parseOpt(k, vStr, optDefs),
+		})
+
+	} else {
+		for j < len(k) {
+			flag := k[j : j+1]
+
+			if od, ok := optDefs[flag]; !ok {
```
whyrusleeping: instead of a big long if-else, the 'idiomatic go' way of doing this is just to have a switch statement

keks: good point!

cli/parse.go
keks: no good reason. I put it into the for header now.

cli/parse_test.go
```diff
@@ -178,28 +179,6 @@ func TestArgumentParsing(t *testing.T) {
 					cmdkit.StringArg("b", true, false, "another arg"),
 				},
 			},
-			"stdinenabled": {
```
whyrusleeping: dropping tests scares me

cli/parse_test.go
```diff
@@ -254,6 +234,55 @@ func TestArgumentParsing(t *testing.T) {
...
+func TestBodyArgs(t *testing.T) {
+	rootCmd := &cmds.Command{
+		Subcommands: map[string]*cmds.Command{
+			"noarg": {},
+			"stdinenabled": {
+				Arguments: []cmdkit.Argument{
+					cmdkit.StringArg("a", true, true, "some arg").EnableStdin(),
+				},
+			},
+			"stdinenabled2args": &cmds.Command{
```
whyrusleeping: oh, here they are

cli/parse_test.go
```diff
@@ -107,41 +109,40 @@ func TestOptionParsing(t *testing.T) {
...
+	test("-b test false", kvs{"bool": true}, words{"false"})
+	test("-b --string foo test bar", kvs{"bool": true, "string": "foo"}, words{"bar"})
+	test("-b=false --string bar", kvs{"bool": false, "string": "bar"}, words{})
 	testFail("foo test")
+	test("defaults", kvs{"opt": "def"}, words{})
+	test("defaults -o foo", kvs{"opt": "foo"}, words{})
 }
 
 func TestArgumentParsing(t *testing.T) {
-	if runtime.GOOS == "windows" {
```
whyrusleeping: does this mean we have windows stuff working now?????

keks: I, uhm, don't know. Not exactly sure why I deleted that, sorry.

keks: I've put the skip back in for now.

cli/parse.go
keks: alright!

cli/run.go
```diff
@@ -0,0 +1,154 @@
...
+	cmds "github.com/ipfs/go-ipfs-cmds"
+)
+
+type Closer interface {
+	Close()
+}
+
+func Run(ctx context.Context, root *cmds.Command,
+	cmdline []string, stdin, stdout, stderr *os.File,
+	buildEnv func(context.Context, *cmds.Request) (interface{}, error),
```
whyrusleeping: maybe make the buildEnv its own type, as well as the executor

cli/run.go
```diff
@@ -0,0 +1,154 @@
...
+			errCh <- err
+		}
+	}()
+
+	select {
+	case err := <-errCh:
+		printErr(err)
+
+		if kiterr, ok := err.(*cmdkit.Error); ok {
+			err = *kiterr
```
whyrusleeping: why `*kiterr` ?

command.go
```diff
@@ -156,13 +154,16 @@ func (c *Command) Get(path []string) (*Command, error) {
 
 // GetOptions returns the options in the given path of commands
 func (c *Command) GetOptions(path []string) (map[string]cmdkit.Option, error) {
+	if c == nil {
```
whyrusleeping: when would we encounter this?

encoding.go
```diff
@@ -29,7 +29,8 @@ const (
 	Protobuf    = "protobuf"
 	Text        = "text"
 	TextNewline = "textnl"
-	CLI         = "cli"
+
+	CLI = "*cli.responseEmitter"
```
whyrusleeping: what?

cli/run.go
keks: it's a cmd**kit.Err**or.

command.go
keks: Whenever we create a new request with a nil root command. Especially during testing it's handy to have the thing behave normal (or at least doesn't crash), even if you don't initialize every detail.    But yeah, maybe it's better not to allow this to avoid silent errors. I had to add an empty command in some tests after removing the check, but they pass again.

encoding.go
keks: Oh yes, I thought about finding a better way to do this.

Now that cli.Run doesn't know very much about the executor, the only way to check which postrun to use is to look at the ResponseEmitter. This was the simplest way to do it.
I can also add a `type Namer interface { Name() string }`, make `*cli.responseEmitter` implement that to return `"cli"` and change this line back.

Why didn't I have that idea earlier again?

http/client.go
```diff
@@ -146,3 +228,17 @@ func getQuery(req cmds.Request) (string, error) {
 
 	return query.Encode(), nil
 }
+
+func isConnRefused(err error) bool {
```
whyrusleeping: 👏

keks: I'd love to take the praise but I'm pretty sure I copied that from corehttp or so :(

http/config.go
```diff
@@ -0,0 +1,144 @@
...
+	APIPath string
+
+	// Headers is an optional map of headers that is written out.
+	Headers map[string][]string
+
+	// corsOpts is a set of options for CORS headers.
+	corsOpts *cors.Options
+
+	// corsOptsRWMutex is a RWMutex for read/write CORSOpts
+	corsOptsRWMutex sync.RWMutex
```
whyrusleeping: I like this pattern that @Kubuxu pointed out recently: https://play.golang.org/p/ddVfHDYu06Y

http/config.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+
+	return false
+}
+
+// allowReferer this is here to prevent some CSRF attacks that
+// the API would be vulnerable to. We check that the Referer
+// is allowed by CORS Origin (origins and referrers here will
+// work similarly in the normla uses of the API).
```
whyrusleeping: typo: normal

http/handler.go
```diff
@@ -168,20 +142,18 @@ func (i internalHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	reqLogEnt := i.ctx.ReqLog.Add(req)
-	defer i.ctx.ReqLog.Finish(reqLogEnt)
-
-	//ps: take note of the name clash - commands.Context != context.Context
-	req.SetInvocContext(i.ctx)
+	if req.Context == nil {
+		fmt.Println("cmd.Call: request nil")
```
whyrusleeping: request is nil? or request context is nil?

http/handler.go
```diff
@@ -168,20 +142,18 @@ func (i internalHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		debug.PrintStack()
```
whyrusleeping: after we print the stack here, we want to continue? or do we want to halt?

http/handler.go
```diff
@@ -168,20 +142,18 @@ func (i internalHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
 
-	err = req.SetRootContext(ctx)
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+	if reqLogger, ok := h.env.(requestLogger); ok {
```
whyrusleeping: 👏 👏

http/handler.go
keks: Sorry, that was an accident. Ignore the whole if block.

http/config.go
keks: Yes, that's neat. I don't take authorship for this part though, I only moved it from handler.go :)

http/parse.go
```diff
@@ -1,42 +1,38 @@
...
 	cmds "github.com/ipfs/go-ipfs-cmds"
-	path "github.com/ipfs/go-ipfs/path"
 )
 
-// Parse parses the data in a http.Request and returns a command Request object
-func Parse(r *http.Request, root *cmds.Command) (cmds.Request, error) {
-	if !strings.HasPrefix(r.URL.Path, ApiPath) {
-		return nil, errors.New("Unexpected path prefix")
+// parseRequest parses the data in a http.Request and returns a command Request object
+func parseRequest(ctx context.Context, r *http.Request, root *cmds.Command) (*cmds.Request, error) {
```
whyrusleeping: Why unexport this?

http/parse.go
```diff
@@ -136,7 +144,8 @@ func Parse(r *http.Request, root *cmds.Command) (cmds.Request, error) {
 		return nil, err
 	}
 
-	return req, nil
+	err = req.FillDefaults()
+	return req, err
```
whyrusleeping: So would we want to not return nil for the request if FillDefaults fails?

keks: No! We can't do that, because we want to print the help text if it was requested, even if some of the parsing failed. For that we need to return the incomplete request.

http/parse.go
keks: Because it didn't make sense to me why it should be exported (especially since parseResponse was not exported)

http/parse.go
```diff
@@ -148,16 +157,70 @@ func parseOptions(r *http.Request) (map[string]interface{}, []string) {
...
-	_, short := opts[cmdkit.EncShort]
-	_, long := opts[cmdkit.EncLong]
-	if !short && !long {
-		opts[cmdkit.EncShort] = cmds.JSON
+	return opts, args
+}
+
+// parseResponse decodes a http.Response to create a cmds.Response
+func parseResponse(httpRes *http.Response, req *cmds.Request) (cmds.Response, error) {
+	var err error
```
whyrusleeping: this seems unnecessary

request.go
```diff
@@ -3,376 +3,145 @@ package cmds
...
-
-func (r *request) Files() files.File {
-	return r.files
+	return bufio.NewScanner(fi), nil
 }
 
-func (r *request) SetFiles(f files.File) {
-	r.files = f
+func IsAllArgsAlreadyCovered(err error) bool {
+	return err != nil && err.Error() == "all arguments covered by positional arguments"
```
whyrusleeping: oooooo, can we try to avoid checking errors by their string?

request.go
```diff
@@ -3,376 +3,145 @@ package cmds
...
-	if optVal := req.Option("timeout"); optVal != nil {
-		var err error
-		tout, found, err = optVal.String()
-		if err != nil {
-			return nil, fmt.Errorf("error parsing timeout option: %s", err)
-		}
-	}
+func (req *Request) SetOption(name string, value interface{}) {
+	optDefs, err := req.Root.GetOptions(req.Path)
+	optDef, found := optDefs[name]
```
whyrusleeping: wait, if GetOptions fails, does it not return nil? This would panic in that case, unless i'm missing something

request.go
```diff
@@ -381,73 +150,54 @@ func (r *request) ConvertOptions() error {
...
-		path:       path,
-		options:    opts,
-		arguments:  args,
-		files:      file,
-		cmd:        cmd,
-		ctx:        ctx,
-		optionDefs: optDefs,
-		values:     values,
-		stdin:      os.Stdin,
+	switch enc := encIface.(type) {
```
whyrusleeping: seems a bit hacky, whats the issue here? something about marshaling back and forth not preserving string types?

request.go
keks: It returns nil, but reading a nil map gives you (<zero>, false): https://play.golang.org/p/vQMkk4nuVN4

request.go
keks: yeah, that was hacky. using type assertions on the error now.

request.go
keks: yes, cmdkit.Option does not know about types, only about reflect.Kind (and that is reflect.String for both type string and EncodingType). I think in the long run cmdkit should be taught to handle types properly, especially for that strings PR that has been out there for a while now. It assumes "if kind is Array, it's got to be a string array!" which is wrong imho, but doing it right would require a lot of changes and it didn't have priority.

request.go
whyrusleeping: whoa what. TIL

single_test.go
```diff
@@ -1,12 +1,13 @@
 package cmds
 
 import (
+	"context"
 	"io"
 	"testing"
 )
 
 func TestSingle_1(t *testing.T) {
-	req, err := NewRequest(nil, nil, nil, nil, nil, nil)
```
Stebalien: I'd use an actual context here (and cancel it after the test).

http/responseemitter.go
```diff
@@ -175,11 +184,8 @@ func (re *responseEmitter) Flush() {
 }
 
 func (re *responseEmitter) preamble(value interface{}) {
-	h := re.w.Header()
-	// Expose our agent to allow identification
-	h.Set("Server", "go-ipfs/"+config.CurrentVersionNumber)
```
Stebalien: Any reason to get rid of this?

http/client.go
```diff
@@ -88,9 +159,18 @@ func (c *client) Send(req cmds.Request) (cmds.Response, error) {
 	} else {
 		httpReq.Header.Set(contentTypeHeader, applicationOctetStream)
 	}
-	httpReq.Header.Set(uaHeader, config.ApiVersion)
```
Stebalien: Are we doing this somewhere else?

http/client.go
```diff
@@ -88,9 +159,18 @@ func (c *client) Send(req cmds.Request) (cmds.Response, error) {
...
+	httpReq.Header.Set(uaHeader, c.ua)
+
+	var reqCancel func()
+	if timeoutStr, ok := req.Options[cmds.TimeoutOpt]; ok {
+		timeout, err := time.ParseDuration(timeoutStr.(string))
```
Stebalien: We should have a `DurationOption`. Also, see #50. Thoughts?

http/client.go
```diff
@@ -88,9 +159,18 @@ func (c *client) Send(req cmds.Request) (cmds.Response, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		req.Context, reqCancel = context.WithTimeout(req.Context, timeout)
+	}
 
-	httpReq.Cancel = req.Context().Done()
+	httpReq.Cancel = req.Context.Done()
```
Stebalien: While we're at it, I believe the correct way to do this now is to call `httpReq.WithContext(req.Context)`.

http/client.go
```diff
@@ -26,40 +30,107 @@ var OptionSkipMap = map[string]bool{
...
+		return err
 	}
 
+	return cmds.Copy(re, res)
+}
+
+func (c *client) Send(req *cmds.Request) (cmds.Response, error) {
+	if req.Context == nil {
+		log.Warningf("no context set in request")
+		req.Context = context.TODO()
```
Stebalien: This should be the background context.

cli/parse.go
```diff
@@ -14,265 +15,186 @@ import (
...
 )
 
 var log = logging.Logger("cmds/cli")
 
 // Parse parses the input commandline string (cmd, flags, and args).
 // returns the corresponding command Request object.
-func Parse(input []string, stdin *os.File, root *cmds.Command) (cmds.Request, *cmds.Command, []string, error) {
-	path, opts, stringVals, cmd, err := parseOpts(input, root)
+func Parse(ctx context.Context, input []string, stdin *os.File, root *cmds.Command) (*cmds.Request, error) {
+	req, err := parse(input, root)
```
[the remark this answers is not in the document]
whyrusleeping: @Stebalien how do you think we should change it?

http/client.go
whyrusleeping: Yeah, we now pass in the userAgent when constructing the client, and then set it roughly around here

http/responseemitter.go
whyrusleeping: We now do this by adding it to the Headers in the ServerConfig. Since this lib isnt in go-ipfs (and shouldnt depend on it anymore) we cant import `config`

http/client.go
whyrusleeping: gonna punt on that until later, but I agree

cli/parse.go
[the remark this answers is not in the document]
Stebalien: Not sure, this just rubs me the wrong way. I'll take a crack at refactoring it and see if I can make it any cleaner.

http/client.go
Stebalien: Got it.

http/handler.go
```diff
@@ -191,17 +157,17 @@ func (i internalHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		re.SetError(err, cmdkit.ErrNormal)
 		return
 	}
 
 	d := re.(Doner)
 	select {
 	case <-d.Done():
-	case <-req.Context().Done():
-		log.Errorf("waiting for http.Responseemitter to close but then %s", req.Context().Err())
+	case <-req.Context.Done():
```
Stebalien: Should we not also be waiting on `r.Context()` (the HTTP request context)?

cli/run.go
```diff
@@ -0,0 +1,153 @@
...
+		re, exitCh = NewResponseEmitter(stdout, stderr, enc, req)
+	} else {
+		return fmt.Errorf("could not find matching encoder for enctype %#v", encType)
+	}
+
+	errCh := make(chan error)
+	go func() {
+		err := exctr.Execute(req, re, env)
+		if err != nil {
+			errCh <- err
```
Stebalien: Goroutine leak. Need to allocate a size-one buffer.

Stebalien:
@whyrusleeping I made a few more changes to try to get the tests green. The only one I'm unsure of is https://github.com/ipfs/go-ipfs-cmds/pull/49/commits/19525e62df326099aa5d81c7e847ede2b90a8d6a#diff-58bccd1b2cfaf06d4d4ac893e4712c88R81 I assume the request is over once the executor returns, right?

whyrusleeping:
@Stebalien LGTM, lets ship this

PR #51, opened 2018-01-09
remove context from the environment and use the context from the request

Stebalien:


Stebalien:
Nevermind. This patch was incorrect (we *do* need to wait on that CloseNotify channel).

PR #54, opened 2018-01-11
[RFC] don't generate a stack trace for every command

Stebalien:
We can add these back if we need them but, at the very least, we shouldn't be unconditionally generating stack traces on `Close`.

### DIFF
--- a/cli/responseemitter.go
+++ b/cli/responseemitter.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"runtime/debug"
 	"sync"
 
 	"github.com/ipfs/go-ipfs-cmdkit"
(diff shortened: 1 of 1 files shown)


cli/responseemitter.go
```diff
@@ -172,7 +169,6 @@ func (re *responseEmitter) Emit(v interface{}) error {
 	}
 
 	if err, ok := v.(cmdkit.Error); ok {
-		log.Warningf("fixerr %s", debug.Stack())
```
Stebalien: What was this fixme for? Not calling `Emit(err)`?

cli/responseemitter.go
```diff
@@ -88,15 +87,13 @@ func (re *responseEmitter) isClosed() bool {
...
-
-	log.Debugf("err=%v exit=%v\nStack:\n%s", re.err, re.exit, debug.Stack())
-
 	re.wLock.Lock()
 	defer re.wLock.Unlock()
 
+	if re.closed {
+		return errors.New("closing closed responseemitter")
+	}
+
```
Stebalien: Because we might as well do this while we're at it.

PR #63, opened 2018-02-09
Better arrays

whyrusleeping:
It was previously pretty awkward to have to use a pointer to an array.


whyrusleeping:
Yeah, it allows using slices directly instead of pointers to slices.

PR #64, opened 2018-02-14
return errors from cmds.Copy directly

whyrusleeping:
Fixes the issue seen in https://github.com/ipfs/go-ipfs/issues/4683 But i'm not 100% sure this is the right way.

cc @kevina, you mentioned you could help look into cmds lib issues.

### DIFF
--- a/responseemitter.go
+++ b/responseemitter.go
@@ -72,6 +72,10 @@ func Copy(re ResponseEmitter, res Response) error {
 			return err
 		}
 
+		if err, ok := v.(cmdkit.Error); ok {
+			return err
+		}
+
 		err = re.Emit(v)
 		if err != nil {
 			return err


kevina:
This might be a problem if multiple errors are emitted, but I am not sure yet.  Right now I don't think we do it and it could be that emitting multiple errors was part of the original design, but only @keks can know for sure.

Stebalien:
As far as I can tell, this code is already doing what it's supposed to do. `Response.Next()` is supposed to determine that the value is actually an error and return it as an error.

Stebalien:
I believe the correct fix is #66. The other `Response` implementations already do this.

kevina:
This changes the output of the `repo gc` command that does essentially emit two errors.  Before `t0087-repo-robust-gc.sh` `47 - 'ipfs repo gc' should abort with two errors` the output would be:
```
Error: could not retrieve links for QmSijovevteoY63Uj1uC5b8pkpDU5Jgyk2dYBqz3sMJUPc: merkledag: not found
Error: could not retrieve links for QmTbPEyrA1JyGUHFvmtx1FNZVzdBreMv8Hc8jV9sBRWhNA: The block referred to by 'QmTbPEyrA1JyGUHFvmtx1FNZVzdBreMv8Hc8jV9sBRWhNA' was not a valid merkledag node
Error: encountered errors during gc run
Error: garbage collection aborted: could not retrieve some links
```
Now it is just
```Error: could not retrieve links for QmSijovevteoY63Uj1uC5b8pkpDU5Jgyk2dYBqz3sMJUPc: merkledag: not found
Error: could not retrieve links for QmTbPEyrA1JyGUHFvmtx1FNZVzdBreMv8Hc8jV9sBRWhNA: The block referred to by 'QmTbPEyrA1JyGUHFvmtx1FNZVzdBreMv8Hc8jV9sBRWhNA' was not a valid merkledag node
Error: encountered errors during gc run
```
The last error not emitted is the error returned from the GC() function, which is concerning.

This needs to sorted out.

PR #66, opened 2018-02-16
handle both cmdkit.Error and *cmdkit.Error in ChanResponse

Stebalien:
fixes https://github.com/ipfs/go-ipfs/issues/4683


kevina:
Assuming this fixes the problem we are having this LGTM.

PR #67, opened 2018-02-16
cleaner command completion/calling

Stebalien:
**Returning errors/calling:**

Instead of having a dance where we:
  * Close the response emitter immediately if the command succeeds.
  * Return an error, call SetError, and then close the response emitter if it fails.

Just set the error from within the `Call` command and always close the response emitter.

**Doner**

Get rid of the Done channel/doner interface from the response emitter. We close the response emitter after finishing the Call *anyways* so this channel isn't necessary (anymore, at least).

### DIFF
--- a/command.go
+++ b/command.go
@@ -66,27 +66,26 @@ var ErrNoFormatter = ClientError("This command cannot be formatted to plain text
 var ErrIncorrectType = errors.New("The command returned a value with a different type than expected")
 
 // Call invokes the command for the given Request
-func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) (err error) {
+func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) {
 	// we need the named return parameter so we can change the value from defer()
-	defer func() {
-		if err == nil {
-			re.Close()
-		}
-	}()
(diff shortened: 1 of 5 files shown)


command.go
```diff
@@ -104,23 +103,7 @@ func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) (err e
 		}
 	}
 
-	defer func() {
```
Stebalien: Dead code. I'm not a fan of recovering from panics anyways (usually). We have no guarantees about the current state at that point.

http/handler.go
```diff
@@ -158,24 +157,7 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	err = h.root.Call(req, re, h.env)
-	if err != nil {
-		re.SetError(err, cmdkit.ErrNormal)
-		return
-	}
-
-	d := re.(Doner)
-	select {
-	case <-d.Done():
-	case <-req.Context.Done():
```
Stebalien: The command waits on this internally anyways.

whyrusleeping:
Alright, LGTM.

PR #68, opened 2018-02-28
fix: improve cli/run to return errors and exit code

dignifiedquire:
Before this was not really usable as errors were printed `Error: <error>` and the exitcode was printed instead of returned.

### DIFF
--- a/cli/run.go
+++ b/cli/run.go
@@ -12,6 +12,13 @@ import (
 	cmds "github.com/ipfs/go-ipfs-cmds"
 )
 
+// ExitError is the error used when a specific exit code needs to be returned.
+type ExitError int
+
+func (e ExitError) Error() string {
+	return fmt.Sprintf("exit code %d", int(e))
+}
+
 // Closer is a helper interface to check if the env supports closing
 type Closer interface {
 	Close()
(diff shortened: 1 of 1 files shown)


whyrusleeping:
@dignifiedquire it looks like the simple way to do this change would have just been to export the ExitErr type.

dignifiedquire:
hmm maybe, not sure having to match on the error type to extract a potential exit code is a good interface though

whyrusleeping:
I think it makes a bit more sense though, because the exit codes you are returning in other places are nonsensical (not actual exit codes, just standins returned by the library). Theres no good reason to restrict ourselves to an integer to express *why* failure happened unless we really need to.

dignifiedquire:
> to restrict ourselves to an integer to express why failure happened

Correct, which is why it returns both the error and the code

whyrusleeping:
right, but the code means nothing in 99% of the cases youre returning it. The only time that error code is relevant is at the end, where we are returning the exitErr

dignifiedquire:
What I would see as a potential better solution is to extend `ExitError` to contain both a code and an error, such that it wraps around other errors, with the added benefit of having a code attached, i.e. 

```go
type ExitError struct {
  err error
  Code int
}

func (e *ExitError) Error() error {
  return e.err
}
```

whyrusleeping:
How is that better? At no point do we return both an error and an exit code.

dignifiedquire:
true, I guess what we actually want to return is a union, but we can't do that so I'll revert to using the old exitError :/

PR #74, opened 2018-03-13
allow stdin on Windows

djdv:
This may be all that has to be done in cmds for now, the rest may be in go-ipfs itself.
Ref: https://github.com/ipfs/go-ipfs/issues/4808

### DIFF
--- a/cli/parse.go
+++ b/cli/parse.go
@@ -7,18 +7,25 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"runtime"
 	"sort"
 	"strings"
 
 	"github.com/ipfs/go-ipfs-cmdkit"
 	"github.com/ipfs/go-ipfs-cmdkit/files"
 	"github.com/ipfs/go-ipfs-cmds"
-
 	logging "github.com/ipfs/go-log"
+
+	osh "gx/ipfs/QmXuBJ7DR6k3rmUEKtvVMhwjmXDuJgXXPUt4LQXKBMsU93/go-os-helper"
 )
 
 var log = logging.Logger("cmds/cli")
+var msgStdinInfo = "ipfs: Reading from %s; send Ctrl-d to stop."
+
+func init() {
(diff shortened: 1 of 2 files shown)


djdv:
FWIW I plan on adding something similar to this in `go-ipfs/shareness`
```
  test_expect_success "stdin reading message shows up" '
  case $(uname) in
  MSYS*)
    test_expect_code 1 grep "ipfs: Reading from" add_help_err &&
    test_expect_code 1 grep "send Ctrl-z to stop." add_help_err
    ;;
  *)
    test_expect_code 1 grep "ipfs: Reading from" add_help_err &&
    test_expect_code 1 grep "send Ctrl-d to stop." add_help_err
    ;;
  esac
  '
```
which should address what codecov is upset over.

Stebalien:
> This may be all that has to be done in cmds for now, the rest may be in go-ipfs itself.

Does this actually fix the issue in question or does something else need to be done to make stdin work on windows? 

Do you have any idea why this check was added? I assume there was a reason.

djdv:
>Does this actually fix the issue in question or does something else need to be done to make stdin work on windows?

Partially, see https://github.com/ipfs/go-ipfs/issues/4808#issuecomment-372678893
stdin+stdout appear to be working as long as the daemon is running, I'll have to investigate `go-ipfs` to see if something can be done about the lock while the daemon is down. I can't see why it's not released before the next command is issued.

I'm going to bounce back and forth trying to get tests around stdio working on Windows.

>Do you have any idea why this check was added? I assume there was a reason.

It's a little unclear unfortunately.
https://github.com/ipfs/go-ipfs/issues/3266

whyrusleeping:
@djdv any update here? Should we just merge this and push it into go-ipfs?

Stebalien:
@whyrusleeping @djdv 

Let's merge it. I still don't know why we disabled this command on Windows in the past (and not Linux) but we might as well try it.

PR #78, opened 2018-03-17
prefetch required stdin arguments in CheckArguments

Stebalien:
Also:

* Make it easier to actually use ParseBodyArguments. Now it only returns an error when something actually goes wrong.
* Add a function for validating a command tree (to be run from a test case).

### DIFF
--- a/command.go
+++ b/command.go
@@ -244,7 +244,7 @@ func (c *Command) CheckArguments(req *Request) error {
 	}
 
 	// iterate over the arg definitions
-	valueIndex := 0 // the index of the current value (in `args`)
+	requiredStringArgs := 0 // number of required string arguments
 	for _, argDef := range req.Command.Arguments {
 		// Is this a string?
 		if argDef.Type != cmdkit.ArgString {
@@ -257,9 +257,10 @@ func (c *Command) CheckArguments(req *Request) error {
 			// Yes, we're all done.
 			break
 		}
+		requiredStringArgs++
 
-		// Do we have this required string argument?
-		if valueIndex < len(req.Arguments) {
+		// Do we have enough string arguments?
+		if requiredStringArgs <= len(req.Arguments) {
 			// all good
 			continue
 		}
--- a/command.go
+++ b/command.go
@@ -225,7 +225,7 @@ func (c *Command) CheckArguments(req *Request) error {
 	}
 
 	lastArg := c.Arguments[len(c.Arguments)-1]
-	if req.bodyArgs == nil &&
+	if req.bodyArgs == nil && // check this as we can end up calling CheckArguments multiple times. See #80.
 		lastArg.SupportsStdin &&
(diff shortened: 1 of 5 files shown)


Stebalien:
(tested with go-ipfs and it fixes the issue in question)

command.go
```diff
@@ -162,55 +164,120 @@ func (c *Command) GetOptions(path []string) (map[string]cmdkit.Option, error) {
...
+		return nil
+	}
 
-	// count required argument definitions
-	numRequired := 0
-	for _, argDef := range c.Arguments {
-		if argDef.Required {
-			numRequired++
+	lastArg := c.Arguments[len(c.Arguments)-1]
+	if req.bodyArgs == nil &&
```
whyrusleeping: maybe add a "TODO: avoid calling this function twice"

Stebalien: Filed an issue.

command.go
```diff
@@ -162,55 +164,121 @@ func (c *Command) GetOptions(path []string) (map[string]cmdkit.Option, error) {
...
+		lastArg.SupportsStdin &&
+		lastArg.Type == cmdkit.ArgString &&
+		req.Files != nil {
+
+		fi, err := req.Files.NextFile()
+		switch err {
+		case io.EOF:
+		case nil:
+			req.bodyArgs = bufio.NewScanner(fi)
+			// Can't pass files and stdin arguments.
```
Kubuxu: This mean that `ipfs add ./file -` is not possible?

command.go
```diff
@@ -162,55 +164,121 @@ func (c *Command) GetOptions(path []string) (map[string]cmdkit.Option, error) {
...
+	if len(c.Arguments) == 0 {
+		return nil
+	}
 
-	// count required argument definitions
-	numRequired := 0
-	for _, argDef := range c.Arguments {
-		if argDef.Required {
-			numRequired++
+	lastArg := c.Arguments[len(c.Arguments)-1]
```
Kubuxu: Won't this panic in case of command with no arguments?

Kubuxu: Ahh, line above.

command.go
Kubuxu: Ok, so this is only ArgString type.

Stebalien: Sorry, stdin *string* arguments.

command.go
```diff
@@ -162,55 +164,121 @@ func (c *Command) GetOptions(path []string) (map[string]cmdkit.Option, error) {
...
+		lastArg.Type == cmdkit.ArgString &&
+		req.Files != nil {
+
+		fi, err := req.Files.NextFile()
+		switch err {
+		case io.EOF:
+		case nil:
+			req.bodyArgs = bufio.NewScanner(fi)
+			// Can't pass files and stdin arguments.
+			req.Files = nil
```
whyrusleeping: I don't think we need to set this to nil here.

It should be fine that we have called NextFile (popping that off the queue)

PR #81, opened 2018-03-17
restore Error: prefix

Stebalien:
this is a breaking change and many of our tests expect this

### DIFF
--- a/cli/run.go
+++ b/cli/run.go
@@ -29,7 +29,7 @@ func Run(ctx context.Context, root *cmds.Command,
 	buildEnv cmds.MakeEnvironment, makeExecutor cmds.MakeExecutor) error {
 
 	printErr := func(err error) {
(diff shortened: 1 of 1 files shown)


dignifiedquire:
any chance we can make that an option? otherwise I will need to maintain a copy with that change :/

Stebalien:
> any chance we can make that an option? otherwise I will need to maintain a copy with that change

I'd rather not; I can't see a clean way of doing that without adding yet another argument to that function. Why do you need this change?

Ideally, we wouldn't print any errors, we'd just return them. However, that makes printing help awkward.

Stebalien:
I'm going to merge this for now as this was a breaking change, we need to cut a go-ipfs release, and I didn't really consider the impact when I merged #68. We can discuss solutions to this in #82.

PR #83, opened 2018-03-18
forward the remaining of the stdin args to the server

Stebalien:


arguments.go
```diff
@@ -0,0 +1,66 @@
+package cmds
```
Stebalien: Need to be able to continue treating the argument reader as an `io.ReadCloser` so we can't use `bufio.Scanner`.

cli/parse_test.go
```diff
@@ -503,8 +503,14 @@ func TestBodyArgs(t *testing.T) {
...
-		for s.Scan() {
-			bodyArgs = append(bodyArgs, s.Text())
+		for {
+			next, err := s.Next()
+			if err == io.EOF {
+				break
+			} else if err != nil {
+				t.Fatal(err)
+			}
+			bodyArgs = append(bodyArgs, next)
```
Stebalien: This interface is a bit awkward but I'm not a fan of the scanner interface either.

http/client.go
```diff
@@ -138,8 +138,14 @@ func (c *client) Send(req *cmds.Request) (cmds.Response, error) {
...
 	var fileReader *files.MultiFileReader
 	var reader io.Reader
-
-	if req.Files != nil {
+	if bodyArgs := req.BodyArgs(); bodyArgs != nil {
+		// In the end, this wraps a file reader in a file reader.
+		// However, such is life.
+		fileReader = files.NewMultiFileReader(files.NewSliceFile("", "", []files.File{
+			files.NewReaderFile("stdin", "", bodyArgs, nil),
+		}), true)
```
Stebalien: Yeah....

Stebalien:
(for the record, I hate this PR)

Stebalien:
> A couple tests would be nice just so we don't have to wait until it gets bubbled up into ipfs to see any regressions

Note: I've already bubbled it to test and it seems to be working (the one issue is a slight change in an error message).

Stebalien:
> I personally prefer the scanner interface, but other than that, this LGTM.

I have little preference so I'll switch back to it.

PR #85, opened 2018-03-19
remove redundant argument check from cli.Parse

Stebalien:
This is also done in the Executor. Unfortunately, doing this *here* causes us to block until we get an argument from stdin, even if we're handling a help call.

I'd rather do all this parsing in one place and clean this logic up but this is the quick fix.


whyrusleeping:
probably want to remove the .gx/post-install file

Stebalien:
So, this PR is crap but it fixes the issue and I can't see a clear way forward to make this suck less so I'm going to leave it as it is. It "works".

Issue #86, opened 2018-03-21
Improve Run function signature to improve error handling


dignifiedquire:
> Pulling this from a comment #62 

It would be really nice if `Run` could return an error which gets just passed to `SetError` if it is set, for the simple error case, the code becomes much more aligned with the way regular error handling is written:

```go
func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) {
  err := doSth()
  if err != nil {
    re.SetError(err, cmdkit.ErrNormal)
    return
  }
  re.Emit("cool")
}
// vs
func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
  err := doSth()
  if err != nil {
    return err
  }
  re.Emit("cool")
  return nil
}
```

dignifiedquire:
cc @clkunzang @Stebalien

Stebalien:
Fixed.

PR #87, opened 2018-03-21
Allow Run to return an error to be emitted.

clkunzang:
This addresses #86 straightforwardly.

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -2,6 +2,7 @@ package cmds
 
 import (
 	"context"
+	"errors"
 	"io"
 	"testing"
 	"time"
@@ -34,7 +35,7 @@ func newTestEmitter(t *testing.T) *testEmitter {
 }
 
 // noop does nothing and can be used as a noop Run function
-func noop(req *Request, re ResponseEmitter, env Environment) {}
+func noop(req *Request, re ResponseEmitter, env Environment) error { return nil }
 
 // writecloser implements io.WriteCloser by embedding
 // an io.Writer and an io.Closer
@@ -281,7 +282,7 @@ func TestPostRun(t *testing.T) {
 
 	for _, tc := range testcases {
 		cmd := &Command{
-			Run: func(req *Request, re ResponseEmitter, env Environment) {
+			Run: func(req *Request, re ResponseEmitter, env Environment) error {
 				re.SetLength(tc.length)
 
 				for _, v := range tc.emit {
@@ -294,6 +295,7 @@ func TestPostRun(t *testing.T) {
 				if err != nil {
 					t.Fatal(err)
 				}
+				return nil
 			},
 			PostRun: PostRunMap{
 				CLI: tc.postRun,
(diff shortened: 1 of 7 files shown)


Stebalien:
There's also a call to Run in `command.go`.

command_test.go
```diff
@@ -401,3 +403,35 @@ func TestCancel(t *testing.T) {
...
 }
+
+type testEmitterWithError testing.T
+
+func (s *testEmitterWithError) Close() error {
+	return nil
+}
+
+func (s *testEmitterWithError) SetLength(_ uint64)                              {}
+func (s *testEmitterWithError) SetError(err interface{}, code cmdkit.ErrorType) {}
```
Stebalien: We should record the error here and make sure it's set before the test ends.

clkunzang: Done.

command_test.go
```diff
@@ -2,10 +2,15 @@ package cmds
 
 import (
 	"context"
+	"errors"
 	"io"
 	"testing"
 	"time"
 
+	"sync"
```
Stebalien: We generally group like imports together. Standard, then local (this repo), then remote (other repos) with newlines between).

command_test.go
```diff
@@ -401,3 +407,57 @@ func TestCancel(t *testing.T) {
...
+
+func (s *testEmitterWithError) Close() error {
+	return nil
+}
+
+func (s *testEmitterWithError) SetLength(_ uint64) {}
+
+func (s *testEmitterWithError) SetError(err interface{}, code cmdkit.ErrorType) {
+	errorCount++
+	fmt.Printf("errorCount: %d\n", errorCount)
```
Stebalien: Tests generally shouldn't use printf.

command_test.go
```diff
@@ -401,3 +407,57 @@ func TestCancel(t *testing.T) {
...
+type testEmitterWithError testing.T
+
+func (s *testEmitterWithError) Close() error {
+	return nil
+}
+
+func (s *testEmitterWithError) SetLength(_ uint64) {}
+
+func (s *testEmitterWithError) SetError(err interface{}, code cmdkit.ErrorType) {
+	errorCount++
```
Stebalien: We *definitely* shouldn't be using a global here. Instead, I'd stash the error in the test emitter (or set a flag or something).

command_test.go
clkunzang: Grrr... Debugging, sorry.

Stebalien:
LGMT. Thanks!

examples/adder/cmd.go
```diff
@@ -29,36 +29,37 @@ var RootCmd = &cmds.Command{
...
-			Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) {
+			Run: func(req *cmds.Request, re cmds.ResponseEmitter, env cmds.Environment) error {
 				sum := 0
 
 				for i, str := range req.Arguments {
 					num, err := strconv.Atoi(str)
 					if err != nil {
 						re.SetError(err, cmdkit.ErrNormal)
-						return
+						return err
```
dignifiedquire: shouldn't this remove the above `SetError` call, as otherwise these get emitted twice?

clkunzang: Good catch. I'll fix those.

whyrusleeping:
Hey guys, it would be great to not push breaking changes haphazardly. I'm debating reverting this...

whyrusleeping:
Yeah, i'm reverting this. Please try again when there is a reasonably clear upgrade path, and this has been tested in go-ipfs.

whyrusleeping:
master is now 6d7f4fc86b88d912f4f836847b819a50d75c11c6

Stebalien:
@whyrusleeping I've already volunteered to fix go-ipfs and this *does* make returning errors *much* easier.

For old commands, the upgrade path is to modify the shim. For new commands, I plan on fixing this with sed and some manual labor (shouldn't be that hard).

I agree this should have been discussed in an issue and I dropped the ball there. However, I *do* remember discussing this with you in the past and didn't here any objections other than "juan had reasons".

whyrusleeping:
@Stebalien I'm okay with this change in general, but making a breaking change to a library like this should be accompanied by any fixes required downstream. Look at it this way, if we find a critical bug that needs fixing immediately in this package before these changes are made, we have to either revert these changes and make the fix, or we have to do a whole bunch of extra changes (and test them) to get that fix out.

This is especially important given how unstable this particular codebase has been lately.

Stebalien:
Ah. Yeah, fair enough.

PR #88, opened 2018-03-22
Remove calls to SetError when returning error from Run.

clkunzang:
Cleanup from #87.

The remaining calls to `SetError` in the adder example should have been removed.


Stebalien:
Oops. Thanks for catching that.

PR #89, opened 2018-03-26
Create Jenkinsfile

VictorBjelkholm:


VictorBjelkholm:
Windows test is failing, `parse_test.go:487: parsing request for cmd ["stdinenabled"]: expected error %!q(<nil>), got "argument \"a\" is required"` to be exact, @djdv might know whats up

Stebalien:
Ah. Well, stdin is disabled on windows. We should probably disable that test on windows until we fix it.

djdv:
@VictorBjelkholm 
Sorry, I missed this mention.

The current state of stdio on Windows is that it seems to works under every condition except 1, piping from ipfs to ipfs while the daemon is down. Everything else should work, including `ipfs|ipfs` when the daemon is up.
Example: https://github.com/ipfs/go-ipfs/issues/4808#issuecomment-372678893

I need to figure out a solution to the local pipe problem before https://github.com/ipfs/go-ipfs-cmds/pull/74 can be merged.

VictorBjelkholm:
This is passing now, fine to merge?

PR #92, opened 2018-03-28
Use existing code when setting error from returned error.

clkunzang:
This PR adds handling for non-`ErrNormal` error types returned from `Run`.

It is unclear to me why 7 commits are showing up in this PR. Presumably it's related to shenanigans with the feat/filecoin branch.

The last commit is a child of the Gx release 1.1.1, which is also the current head of `feat/filecoin`. Those are the effective changes introduced here.


clkunzang:
Done.

PR #97, opened 2018-04-14
Let Run return an error to be emitted

keks:
This addresses #86 and @whyrusleeping's comments in #87.

### DIFF
--- a/command.go
+++ b/command.go
@@ -106,6 +106,12 @@ func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) {
 
 	err = cmd.Run(req, re, env)
 	if err != nil {
+		if cmderr, ok := err.(cmdkit.Error); ok {
+			err = re.Emit(cmderr)
+			if err != nil {
+				log.Error("error emitting cmd error:", err)
+			}
+		}
 		re.SetError(err, cmdkit.ErrNormal)
 	}
 }
--- a/command.go
+++ b/command.go
@@ -25,7 +25,7 @@ var log = logging.Logger("cmds")
 
 // Function is the type of function that Commands use.
 // It reads from the Request, and writes results to the ResponseEmitter.
-type Function func(*Request, ResponseEmitter, Environment)
+type Function func(*Request, ResponseEmitter, Environment) error
 
 // PostRunMap is the map used in Command.PostRun.
 type PostRunMap map[PostRunType]func(*Request, ResponseEmitter) ResponseEmitter
@@ -104,7 +104,10 @@ func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) {
 		}
 	}
 
-	cmd.Run(req, re, env)
+	err = cmd.Run(req, re, env)
+	if err != nil {
+		re.SetError(err, cmdkit.ErrNormal)
+	}
 }
 
 // Resolve returns the subcommands at the given path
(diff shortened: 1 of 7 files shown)


command.go
```diff
@@ -104,7 +104,16 @@ func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) {
...
-	cmd.Run(req, re, env)
+	err = cmd.Run(req, re, env)
+	if err != nil {
+		if cmderr, ok := err.(cmdkit.Error); ok {
+			err = re.Emit(cmderr)
+			if err != nil {
+				log.Error("error emitting cmd error:", err)
+			}
+		}
+		re.SetError(err, cmdkit.ErrNormal)
```
Stebalien: Won't this emit the error twice?

keks: You're right, fixed now

command.go
```diff
@@ -104,7 +104,18 @@ func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) {
 		}
 	}
 
-	cmd.Run(req, re, env)
+	err = cmd.Run(req, re, env)
+	if err != nil {
+		if cmderr, ok := err.(cmdkit.Error); ok {
```
Stebalien: IIRC, we also need to check if this is a pointer to a `cmdkit.Error`. Right?

command.go
```diff
@@ -104,7 +104,18 @@ func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) {
...
+			err = re.Emit(cmderr)
+			if err != nil {
+				log.Error("error %q emitting cmd error %q on request %#v", err, cmderr, req)
```
Stebalien: I'm worried we'll hit this every time a client disconnects. What do you think?

command.go
keks: *sigh* you're right. We should figure something out to make this less of a hassle - see https://github.com/ipfs/go-ipfs-cmdkit/issues/19

command.go
keks: Hmm, a client disconnect shouldn't trigger an error in Run in the first place, so this entire branch shouldn't trigger - or am I missing something?

Stebalien: Client disconnects -> context gets canceled -> Run returns the context cancellation error.

keks: Hmm, but usually the server will write their response and return. What you write only happens if the server doesn't expect the client to exit, which is a rightful error IMHO.
If the Run function sends data and expects that the client might have closed the connection by now, they should not return context cancellation errors.

Stebalien: > If the Run function sends data and expects that the client might have closed the connection by now, they should not return context cancellation errors.

Telling whether or not the client has gone away can sometimes be a bit tricky and we definitely don't want to return "success" to the user if we *think* the client has disconnected but it hasn't.

Currently, we *always* return an error if we abort the request due to a client disconnect.

keks: Okay, so nothing changes then, right?

Stebalien: I guess in *this* situation, we're unlikely to return a cmdkit Error so we shouldn't hit this log message.

Although, I still don't think that this is a true error case (for the daemon) as the client could have disconnected after run completed but before we try to send the value back. Really, I think that's the only possible way to hit this log.

Stebalien:
Merged in #112

Issue #73, opened 2018-03-08
Ability to use multiplye types


dignifiedquire:
Sometimes commands have different outputs depending on the flags, but I would still like to get type safety using `MakeTypedEncoder`. So I had the idea to allow for the following interface

```go
var myCmd = &cmds.Command{
  // ... 
  Types: []interface{type1, type2},
  Encoders: cmds.EncoderMap{
    cmds.Text: []Encoder{
      MakeTypedEncoder(func(req *cmds.Request, w io.Writer, t1 type1) error {
        // encode t1
      }),
      MakeTypedEncoder(func(req *cmds.Request, w io.Writer, t2 type2) error {
        // encode t2
      }),
    }
}
```

Where the library would go through and find the correct encoder given a certain type. 

In addition I am seeing instances where I would really like to have default encoders registered for certain types, or for certain interfaces. So with the above a nice addition would be to be able to say register these encoders in a global fashion on `cmds` and then they are used as fallback if no matching typed encoder is found.

```go

cmds.RegisterTypedEncoder(MakeTypedEncoder(func(req *cmds.Request, w io.Writer, t1 type1) error {
  // encode t1
}))
cmds.RegisterTypedEncoder(MakeTypedEncoder(func(req *cmds.Request, w io.Writer, t2 type2) error {
  // encode t2
}))

var myCmd = &cmds.Command{
  // ... 
  Types: []interface{type1, type2},
  // Encoders is not here, but it still works, because the encoders for type 1 and type 2 have been registered before
}
```

keks:
Hey @dignifiedquire, I opened a PR that addresses part of what you stated: #99. However, your suggestion lacks the most tricky part of the problem: How does the Decoder know what type to use?

The cleanest approach IMHO is to give hints in the JSON, such as
```json
{
  "type": "fooType",
  "value": {
    // ...
  }
}
```
This is probably not what you meant, but we already have big problems telling a value and an error apart.

The decoding part is not yet in the PR, but I'm working on it.

keks:
Actually, before I dash ahead, let me know if you want JSON that looks like this.

dignifiedquire:
Generally I think having some type information is a good way to go, but this would mean the json would be very different in the multi case vs the singular case. One hacky way I could see is to try the different types, and use the one that doesn't fail, with the first one to succeed being the one that is used.

keks:
>  use the one that doesn't fail

[...]

PR #99, opened 2018-04-16
[WIP] Add MakeMultiEncoder

keks:
This PR addresses #73.

Note that this simply encodes the values. It does not add any type hints, which is what you might actually want. Let's further discuss this in the issue.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -111,6 +111,39 @@ func MakeTypedEncoder(f interface{}) func(*Request) func(io.Writer) Encoder {
 	})
 }
 
+// MakeMultiEncoder takes pairs of arguments, where the first of the pair is a value that denotes a type and the second denotes an encoder.
+// The resulting encoder then uses the encoder of the pair with matching first element.
+// Example:
+//    e := MakeMultiEncoder(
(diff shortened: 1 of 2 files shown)


encoding.go
```diff
@@ -111,6 +111,39 @@ func MakeTypedEncoder(f interface{}) func(*Request) func(io.Writer) Encoder {
 	})
 }
 
+// MakeMultiEncoder takes pairs of arguments, where the first of the pair is a value that denotes a type and the second denotes an encoder.
+// The resulting encoder then uses the encoder of the pair with matching first element.
+// Example:
+//    e := MakeMultiEncoder(
+//    	"string", MakeTypedEncoder(func(req *Request, w io.Writer, str string) error {
```
dignifiedquire: Shouldn't we be able to use reflection to get the type from the function signature, instead of manually writing the type?

keks:
We do, we could also pass an empty string but this is more descriptive. 

Am 16. April 2018 17:32:23 MESZ schrieb Friedel Ziegelmayer <notifications@github.com>:

encoding.go
keks: We do, we could also pass an empty string but this is more descriptive.

PR #101, opened 2018-04-20
Rezero struct between each decode

acruikshank:
## Problem

When the server returns multiple values, the client decodes the results into the Command.Type struct. It does not reset this struct, so any property set in the first struct and not overridden in the second will appear in the second value as it is emitted by the client. For example the server responds with the following new line delimited JSON:

```JSON
{"a": 1}
{"a": 1, "b": 2}
{"a": 1}
```
for a command whose Type is:

```GO
type CmdType struct {
  a int;
  b int;
}
```

We would expect the client to output:
```JSON
{"a": 1, "b": 0}
{"a": 1, "b": 2}
{"a": 1, "b": 0}
```
but it actually outputs:
```JSON
{"a": 1, "b": 0}
{"a": 1, "b": 2}
{"a": 1, "b": 2}
```
Because `b` isn't set in the last result, it keeps the value `2` from the previous result.

## Solution

The proposed solution uses the reflection library to instantiate new instance of the given type for each pass through decode.

### DIFF
--- a/http/response.go
+++ b/http/response.go
@@ -66,11 +66,13 @@ func (res *Response) RawNext() (interface{}, error) {
 		}
 	}
 
-	valueType := reflect.TypeOf(res.req.Command.Type)
-	if valueType.Kind() == reflect.Ptr {
-		valueType = valueType.Elem()
+	var value interface{}
+	if valueType := reflect.TypeOf(res.req.Command.Type); valueType != nil {
+		if valueType.Kind() == reflect.Ptr {
+			valueType = valueType.Elem()
+		}
+		value = reflect.New(valueType).Interface()
 	}
-	value := reflect.New(valueType).Interface()
 	m := &cmds.MaybeError{Value: value}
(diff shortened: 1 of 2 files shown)


Stebalien:
Uh.. Oops. Thank you for catching that! Mind running `go fmt` on your code to make CI happy?

acruikshank:
Oops, myself. Code is formatted and all tests are passing.

Stebalien:
I fixed the case where no type is specified (no output on success).

Stebalien:
Actually, I just merged (not rebased) master into the filecoin branch. We'll never merge the filecoin branch into go-ipfs (probably either re-do the changes or cherry-pick them) so this'll probably be the easiest way to keep the two branches in sync.

New release: QmUf5GFfV2Be3UtSAPKDVkoRd1TwEBTmx9TSSCFGGjNgdQ

acruikshank:
Thanks!

PR #102, opened 2018-04-23
Change PostRun signature

keks:
This change set changes the signature of PostRun function from
```go
func(*Request, ResponseEmitter) ResponseEmitter
```
to
```go
func(Response, ResponseEmitter) error
```
(Note that `Response` lets the callee access the request as well)

Before, every PostRun function had the same boilerplate: first create a new (Reponse, ResponseEmitter) pair, read the response in a goroutine and write to the responseemitter we were passed, and outside the goroutine return the ResponseEmitter from the pair.

This PR moves all that boilerplate into the cmds library, so the user just writes what used to be inside the goroutine into the PostRun function and, when an error occurs, just returns it.
The core change is in [executor.go](#diff-58bccd1b2cfaf06d4d4ac893e4712c88R84), but I also had to change all the tests and examples.

### DIFF
--- a/command.go
+++ b/command.go
@@ -28,7 +28,7 @@ var log = logging.Logger("cmds")
 type Function func(*Request, ResponseEmitter, Environment)
 
 // PostRunMap is the map used in Command.PostRun.
-type PostRunMap map[PostRunType]func(*Request, ResponseEmitter) ResponseEmitter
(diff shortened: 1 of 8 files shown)


Stebalien:
Merged in #112

Issue #98, opened 2018-04-14
Rename SetError to CloseWithError


keks:
At the moment, we basically only use SetError right before we close the connection, and using it for anything else has turned out to make things much more complicated than I thought.

The method `CloseWithError(error) error` is pretty common in Go, and returns the passed error to the caller of `Read` (or in our case `Next`) on the other side of the stream. Type information can be preserved for runtime-analysis, until we send the error through the http api, where it becomes a cmdkit.Error, if it isn't already.

@Stebalien do you have opinions on this?

keks:
Actually not only rename, but also make it close the connection.

Stebalien:
Sounds reasonable. Although, I assume the only use for this method will be from `PostRun`. `Run` can just return the error.

I wonder if it would be worth it to make PostRun synchronous as well and use the same "just return the error" system.

keks:
> I assume the only use for this method will be from PostRun. Run can just return the error.

Well, Command.Call() also calls Run, so if that returns an error it needs to call CloseWithError to pass the error to the caller of the command.

> I wonder if it would be worth it to make PostRun synchronous as well and use the same "just return the error" system.

I'm not sure what you mean here. Do you want to first call run and capture the output, and then let PostRun postprocess the result? Because we did that before the rewrite and that was terrible.
One example: In `pubsub sub` we need to flush the connection once we successfully set up the subscription to send out the http header - otherwise the client would block until the first message is received. That was not possible with the previous system, we had to send an empty dummy value that was ignored by the recipient. I'd rather not go back to that system.

One more note: Using CloseWithError would clarify our sematics around passing errors and make explicit that fatal errors are returned as errors and status errors are returned as values.

Stebalien:
> Well, Command.Call() also calls Run, so if that returns an error it needs to call CloseWithError to pass the error to the caller of the command.

Ah, I forgot about that.

> I'm not sure what you mean here. Do you want to first call run and capture the output, and then let PostRun postprocess the result?

I mean define PostRun as:

```go
type PostRunFunc = func(req *Request, in Response, out ResponseEmitter) error
```

While users *could* call `CloseWithError`, the correct way would be to return an error. Am I missing something here?

Stebalien:
> Well, it depends. Some errors should go to the connection (requested item not found, etc.) while others should be returned (could not write to connection).

[...]

PR #103, opened 2018-04-23
[WIP] Change SetError to CloseWithError

keks:
This change set drops the `SetError` method on `ResponseEmitter` but introduces `CloseWithError`, as suggested in #98.

It sits on top of the branches of PR #97 and #102 to avoid a huge pile of rebasing later, so let's wait with merging until those are in.


keks:
closed in favour of #106

PR #106, opened 2018-05-03
[WIP] Change SetError to CloseWithError

keks:
This change set drops the `SetError` method on `ResponseEmitter` but introduces `CloseWithError`, as suggested in #98.

It sits on top of the branches of PR #97 and #102 to avoid a huge pile of rebasing later, so let's wait with merging until those are in. The `tmp/prebundle` branch has the changes of these PRs, so we can review the changes of this PR individually.

This needs to be squashed before merging, I'll do that when the review is through.

### DIFF
--- a/chan.go
+++ b/chan.go
@@ -10,7 +10,7 @@ import (
 	"github.com/ipfs/go-ipfs-cmdkit"
 )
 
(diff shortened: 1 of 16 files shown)


Stebalien:
Superseded by #122 (which I am reviewing as a set).

Issue #109, opened 2018-05-07
Log requests from the CLI


schomatis:
Currently requests are only being [logged](https://github.com/ipfs/go-ipfs/blob/master/commands/request.go#L69) in the online mode by the [HTTP handler](https://github.com/ipfs/go-ipfs-cmds/blob/master/http/handler.go#L148). It would be useful to extend this mechanism to cover any IPFS use case, which means logging it also in the offline mode in the [CLI](https://github.com/ipfs/go-ipfs-cmds/blob/master/executor.go#L108).

Normally this wouldn't be necessary as the request log is only used as an informative service where there can be more than one request simultaneously (i.e., in the online mode handling many concurrent HTTP request and not in the offline mode with the CLI handling an only request before shutting down) but the rationale is to have a single point of control for any requests (one or many) before shutting down the IPFS core.

schomatis:
I'm having some trouble with the `gx` workflow (see https://github.com/whyrusleeping/gx/issues/174) so I'm labeling this as blocked until I gain a better understanding of it.

keks:
@schomatis Sorry for only responding now.

I'm not sure, but there may be a misunderstanding here. IPFS can call it's commands either directly from the cli, or send it to the daemon via http. Note that this has nothing to do with online or offline mode. Online/Offline mode is a property of your running daemon, and specifies whether or not your ipfs node talks to other ipfs nodes. A daemon in offline mode will still track your requests, since they come in through the http api. When a requests is executed directly through the cli, tracking the requests doesn't make much sense because there is just that one requests that is handled by the process.

Also note that IIUC the request log is only kept in memory, and only as long as the daemon is running. Since in the cli case (i.e. when there is no daemon) there is no process in the background that could track requests, all of this would need to be restructured.

schomatis:
@keks Thanks for the clarification in the terminology.

> When a requests is executed directly through the cli, tracking the requests doesn't make much sense because there is just that one requests that is handled by the process.

Yes but @whyrusleeping [suggested](https://github.com/ipfs/go-ipfs/issues/4820#issuecomment-376687766) using the request log as an indication of command termination (even if it is the only one) to avoid shutting down the node prematurely and wait for the command.

> Since in the cli case (i.e. when there is no daemon) there is no process in the background that could track requests, all of this would need to be restructured.

[...]

PR #111, opened 2018-05-21
[WIP] Log requests from the CLI

schomatis:
Addresses #109.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -38,6 +38,16 @@ type executor struct {
 	root *Command
 }
 
+// RequestLogger copies the `LogRequest` method from
+// `github.com/ipfs/go-ipfs/commands.Context`.
+// Copied from the `http` package as now the request are logged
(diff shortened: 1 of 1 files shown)


schomatis:
@keks Where would be a good point in the execution flow of a CLI command to test if the command was correctly registered in the request logger?

schomatis:
The `Context` object which contains the request log seems to be available at the [`Request`](https://github.com/ipfs/go-ipfs/blob/139d62409214f2b430edc2bc1af87845f565f6a2/commands/request.go#L106) level but I'm not sure if the `IpfsNode` has a direct access to it (for example to check it during the [`teardown`](https://github.com/ipfs/go-ipfs/blob/139d62409214f2b430edc2bc1af87845f565f6a2/core/core.go#L542) procedure). I'll try to take a closer look at the `ipfs diag cmds` command.

Issue #110, opened 2018-05-14
Remove Response.RawNext


keks:
After moving to `CloseWithError`, we can drop that method, because the entire error handling gets much simpler. Also the `HandleError` function can probably go.

I think this should happen before we actually make it a release because that breaks the API and we'd need to make two major releases in a row. I don't feels strong about this though, what's your opinion @Stebalien?

Stebalien:
Fixed.

PR #112, opened 2018-05-24
Bundle of changes

keks:
This branch bundles a couple of changes so we can test against all of them at once in go-ipfs

- #86 Let cmd.Run return an error (PR #97)
- #98 Remove SetError and add CloseWithError
  - also removes HandleError, RecvdErr and RawNext
  - PRs: #103, #110 
- #102 Change PostRun signature so user don't have to build the chanresponse themselves

### DIFF
--- a/chan.go
+++ b/chan.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"runtime/debug"
 	"sync"
 
 	"github.com/ipfs/go-ipfs-cmdkit"
+	"github.com/ipfs/go-ipfs-cmds/debug"
 )
 
 func NewChanResponsePair(req *Request) (ResponseEmitter, Response) {
@@ -137,19 +137,6 @@ func (r *chanResponse) Next() (interface{}, error) {
 type chanResponseEmitter chanResponse
 
 func (re *chanResponseEmitter) Emit(v interface{}) error {
-	// Initially this library allowed commands to return errors by sending an
-	// error value along a stream. We removed that in favour of CloseWithError,
-	// so we want to make sure we catch situations where some code still uses the
-	// old error emitting semantics.
-	// Also errors may occur both as pointers and as plain values, so we need to
-	// check both.
-	if e, ok := v.(cmdkit.Error); ok {
-		v = &e
-	}
-	if e, ok := v.(*cmdkit.Error); ok {
-		log.Errorf("unexpected error value emitted: %s at\n%s", e.Error(), debug.Stack())
-	}
-
 	// channel emission iteration
 	// TODO maybe remove this and use EmitChan instead of calling Emit directly?
 	if ch, ok := v.(chan interface{}); ok {
@@ -169,6 +156,13 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 
 	if _, ok := v.(Single); ok {
 		defer re.closeWithError(io.EOF)
+
+	// Initially this library allowed commands to return errors by sending an
+	// error value along a stream. We removed that in favour of CloseWithError,
+	// so we want to make sure we catch situations where some code still uses the
+	// old error emitting semantics and _panic_ in those situations.
+	debug.AssertNotError(v)
+
 	}
 
 	ctx := re.req.Context
--- a/chan.go
+++ b/chan.go
@@ -125,10 +125,6 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 	re.wl.Lock()
 	defer re.wl.Unlock()
 
-	if _, ok := v.(Single); ok {
-		defer re.closeWithError(nil)
-	}
-
 	// Initially this library allowed commands to return errors by sending an
 	// error value along a stream. We removed that in favour of CloseWithError,
 	// so we want to make sure we catch situations where some code still uses the
@@ -155,6 +151,10 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 
 	select {
 	case re.ch <- v:
+		if _, ok := v.(Single); ok {
+			re.closeWithError(nil)
+		}
+
 		return nil
 	case <-ctx.Done():
 		return ctx.Err()
--- a/chan.go
+++ b/chan.go
@@ -41,8 +41,8 @@ type chanStream struct {
 	closed bool
 
 	// waitErr is closed when the stream is closed.
-	// Error waits for waitErr to be closed.
-	// It is protected by wl.
+	// Error checks if the stream has been closed by checking if this channes is closed.
+	// Its closing is protected by wl.
 	waitErr chan struct{}
 
 	// err is the error that the stream was closed with.
@@ -51,7 +51,7 @@ type chanStream struct {
 
 	// waitLen is closed when the first value is emitted or the stream is closed.
 	// Length waits for waitLen to be closed.
-	// It is protected by wl.
+	// Its closing is protected by wl.
 	waitLen chan struct{}
 
 	// length is the length of the response.
@@ -66,17 +66,20 @@ func (r *chanResponse) Request() *Request {
 }
 
 func (r *chanResponse) Error() *cmdkit.Error {
-	<-r.waitErr
+	select {
+	case <-r.waitErr:
+		if r.err == nil || r.err == io.EOF {
+			return nil
+		}
 
-	if r.err == nil || r.err == io.EOF {
+		if e, ok := r.err.(*cmdkit.Error); ok {
+			return e
+		}
+
+		return &cmdkit.Error{Message: r.err.Error()}
+	default:
 		return nil
 	}
-
-	if e, ok := r.err.(*cmdkit.Error); ok {
-		return e
-	}
-
-	return &cmdkit.Error{Message: r.err.Error()}
 }
 
 func (r *chanResponse) Length() uint64 {
--- a/chan.go
+++ b/chan.go
@@ -116,11 +116,7 @@ func (r *chanResponse) Next() (interface{}, error) {
 	case v, ok := <-r.ch:
 		if !ok {
 			r.ch = nil
-			if r.err != nil {
-				return nil, r.err
-			}
-
-			return nil, io.EOF
+			return nil, r.err
 		}
 
 		if err, ok := v.(*cmdkit.Error); ok {
@@ -225,6 +221,10 @@ func (re *chanResponseEmitter) SetLength(l uint64) {
 	re.length = l
 }
 
+func (re *chanResponseEmitter) Close() error {
+	return re.CloseWithError(nil)
+}
+
 func (re *chanResponseEmitter) CloseWithError(err error) error {
 	re.wl.Lock()
 	defer re.wl.Unlock()
@@ -233,6 +233,10 @@ func (re *chanResponseEmitter) CloseWithError(err error) error {
 }
 
 func (re *chanResponseEmitter) closeWithError(err error) error {
+	if err == nil {
+		err = io.EOF
+	}
+
 	if e, ok := err.(cmdkit.Error); ok {
 		err = &e
 	}
@@ -249,11 +253,4 @@ func (re *chanResponseEmitter) closeWithError(err error) error {
 	})
 
 	return nil
-}
-
-func (re *chanResponseEmitter) Close() error {
-	re.wl.Lock()
-	defer re.wl.Unlock()
-
-	return re.closeWithError(io.EOF)
 }
--- a/chan.go
+++ b/chan.go
@@ -119,7 +119,7 @@ func (r *chanResponse) Next() (interface{}, error) {
 			return nil, r.err
 		}
 
-		if err, ok := v.(*cmdkit.Error); ok {
+		if err, ok := v.(cmdkit.Error); ok {
 			v = &err
 		}
 
--- a/chan.go
+++ b/chan.go
@@ -132,32 +132,6 @@ func (r *chanResponse) Next() (interface{}, error) {
 	case <-ctx.Done():
 		return nil, ctx.Err()
 	}
-
-}
-
-func (r *chanResponse) RawNext() (interface{}, error) {
-	if r == nil {
-		return nil, io.EOF
-	}
-
-	var ctx context.Context
-	if rctx := r.req.Context; rctx != nil {
-		ctx = rctx
-	} else {
-		ctx = context.Background()
-	}
-
-	select {
-	case v, ok := <-r.ch:
-		if ok {
-			return v, nil
-		}
-
-		return nil, io.EOF
-	case <-ctx.Done():
-		return nil, ctx.Err()
-	}
-
 }
 
 type chanResponseEmitter chanResponse
--- a/chan.go
+++ b/chan.go
@@ -27,7 +27,8 @@ func NewChanResponsePair(req *Request) (ResponseEmitter, Response) {
 }
 
 type chanResponse struct {
-	l   sync.Mutex
+	wl  sync.Mutex // lock for writing calls, i.e. Emit et al.
+	rl  sync.Mutex // lock for reading calls, i.e. Next
 	req *Request
 
 	// wait makes header requests block until the body is sent
@@ -104,6 +105,10 @@ func (r *chanResponse) Next() (interface{}, error) {
 	} else {
 		ctx = context.Background()
 	}
+
+	// to avoid races by setting r.ch to nil
+	r.rl.Lock()
+	defer r.rl.Unlock()
 
 	select {
 	case <-r.closed:
@@ -189,12 +194,8 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 		close(re.wait)
 	})
 
-	re.l.Lock()
-	defer re.l.Unlock()
-
-	if re.ch == nil {
-		return fmt.Errorf("emitter closed")
-	}
+	re.wl.Lock()
+	defer re.wl.Unlock()
 
 	if _, ok := v.(Single); ok {
 		defer re.closeWithError(io.EOF)
@@ -203,6 +204,8 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 	ctx := re.req.Context
 
 	select {
+	case <-re.closed:
+		return fmt.Errorf("emitter closed")
 	case re.ch <- v:
 		return nil
 	case <-ctx.Done():
@@ -211,8 +214,8 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 }
 
 func (re *chanResponseEmitter) SetLength(l uint64) {
-	re.l.Lock()
-	defer re.l.Unlock()
+	re.wl.Lock()
+	defer re.wl.Unlock()
 
 	// don't change value after emitting
 	if re.emitted {
@@ -223,8 +226,8 @@ func (re *chanResponseEmitter) SetLength(l uint64) {
 }
 
 func (re *chanResponseEmitter) CloseWithError(err error) error {
-	re.l.Lock()
-	defer re.l.Unlock()
+	re.wl.Lock()
+	defer re.wl.Unlock()
 
 	return re.closeWithError(err)
 }
@@ -249,8 +252,8 @@ func (re *chanResponseEmitter) closeWithError(err error) error {
 }
 
 func (re *chanResponseEmitter) Close() error {
-	re.l.Lock()
-	defer re.l.Unlock()
+	re.wl.Lock()
+	defer re.wl.Unlock()
 
 	return re.closeWithError(io.EOF)
 }
--- a/chan.go
+++ b/chan.go
@@ -2,7 +2,7 @@ package cmds
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"io"
 	"sync"
 
@@ -15,10 +15,9 @@ func NewChanResponsePair(req *Request) (ResponseEmitter, Response) {
 	wait := make(chan struct{})
 
 	r := &chanResponse{
-		req:    req,
-		ch:     ch,
-		wait:   wait,
-		closed: make(chan struct{}),
+		req:  req,
+		ch:   ch,
+		wait: wait,
 	}
 
 	re := (*chanResponseEmitter)(r)
@@ -26,32 +25,44 @@ func NewChanResponsePair(req *Request) (ResponseEmitter, Response) {
 	return re, r
 }
 
-type chanResponse struct {
-	wl  sync.Mutex // lock for writing calls, i.e. Emit et al.
-	rl  sync.Mutex // lock for reading calls, i.e. Next
+// chanStream is the struct of both the Response and ResponseEmitter.
+// The methods are defined on chanResponse and chanResponseEmitter, which are
+// just type definitions on chanStream.
+type chanStream struct {
 	req *Request
 
-	// wait makes header requests block until the body is sent
-	wait chan struct{}
-	// waitOnce makes sure we only close wait once
-	waitOnce sync.Once
+	// rl is a lock for reading calls, i.e. Next.
+	rl sync.Mutex
 
-	// ch is used to send values from emitter to response
+	// ch is used to send values from emitter to response.
+	// When Emit received a channel close, it sets it to nil.
+	// It is protected by rl.
 	ch chan interface{}
 
-	emitted bool
-	err     error
-	length  uint64
+	// wl is a lock for writing calls, i.e. Emit, Close(WithError) and SetLength.
+	wl sync.Mutex
 
-	closeOnce sync.Once
-	closed    chan struct{}
+	// closed stores whether this stream is closed.
+	// It is protected by wl.
+	closed bool
+
+	// wait is closed when the stream is closed or the first value is emitted.
+	// Error and Length both wait for wait to be closed.
+	// It is protected by wl.
+	wait chan struct{}
+
+	// err is the error that the stream was closed with.
+	// It is written once under lock wl, but only read after wait is closed (which also happens under wl)
+	err error
+
+	// length is the length of the response.
+	// It can be set by calling SetLength, but only before the first call to Emit, Close or CloseWithError.
+	length uint64
 }
 
-func (r *chanResponse) Request() *Request {
-	if r == nil {
-		return nil
-	}
+type chanResponse chanStream
 
+func (r *chanResponse) Request() *Request {
 	return r.req
 }
 
@@ -75,25 +86,6 @@ func (r *chanResponse) Length() uint64 {
 	return r.length
 }
 
-func (re *chanResponse) Head() Head {
-	<-re.wait
-
-	var err error
-	if re.err != io.EOF {
-		err = re.err
-	}
-
-	cmdErr, ok := err.(*cmdkit.Error)
-	if !ok && err != nil {
-		cmdErr = &cmdkit.Error{Message: err.Error()}
-	}
-
-	return Head{
-		Len: re.length,
-		Err: cmdErr,
-	}
-}
-
 func (r *chanResponse) Next() (interface{}, error) {
 	if r == nil {
 		return nil, io.EOF
@@ -111,16 +103,9 @@ func (r *chanResponse) Next() (interface{}, error) {
 	defer r.rl.Unlock()
 
 	select {
-	case <-r.closed:
-		return nil, r.err
 	case v, ok := <-r.ch:
 		if !ok {
-			r.ch = nil
 			return nil, r.err
-		}
-
-		if err, ok := v.(cmdkit.Error); ok {
-			v = &err
 		}
 
 		switch val := v.(type) {
@@ -138,7 +123,6 @@ type chanResponseEmitter chanResponse
 
 func (re *chanResponseEmitter) Emit(v interface{}) error {
 	// channel emission iteration
-	// TODO maybe remove this and use EmitChan instead of calling Emit directly?
 	if ch, ok := v.(chan interface{}); ok {
 		v = (<-chan interface{})(ch)
 	}
@@ -146,16 +130,12 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 		return EmitChan(re, ch)
 	}
 
-	// unblock Length(), Error() and Head()
-	re.waitOnce.Do(func() {
-		close(re.wait)
-	})
-
 	re.wl.Lock()
 	defer re.wl.Unlock()
 
 	if _, ok := v.(Single); ok {
-		defer re.closeWithError(io.EOF)
+		defer re.closeWithError(nil)
+	}
 
 	// Initially this library allowed commands to return errors by sending an
 	// error value along a stream. We removed that in favour of CloseWithError,
@@ -163,13 +143,25 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 	// old error emitting semantics and _panic_ in those situations.
 	debug.AssertNotError(v)
 
+	// unblock Length() and Error()
+	select {
+	case <-re.wait:
+	default:
+		close(re.wait)
+	}
+
+	// make sure we check whether the stream is closed *before accessing re.ch*!
+	// re.ch is set to nil, but is not protected by a shared mutex (because that
+	// wouldn't make sense).
+	// re.closed is set in a critical section protected by re.wl (we also took
+	// that lock), so we can be sure that this check is not racy.
+	if re.closed {
+		return ErrClosedEmitter
 	}
 
 	ctx := re.req.Context
 
 	select {
-	case <-re.closed:
-		return fmt.Errorf("emitter closed")
 	case re.ch <- v:
 		return nil
 	case <-ctx.Done():
@@ -177,30 +169,37 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 	}
 }
 
+func (re *chanResponseEmitter) Close() error {
+	return re.CloseWithError(nil)
+}
+
 func (re *chanResponseEmitter) SetLength(l uint64) {
 	re.wl.Lock()
 	defer re.wl.Unlock()
 
-	// don't change value after emitting
-	if re.emitted {
-		return
+	// don't change value after emitting or closing
+	select {
+	case <-re.wait:
+	default:
+		re.length = l
 	}
-
-	re.length = l
-}
-
-func (re *chanResponseEmitter) Close() error {
-	return re.CloseWithError(nil)
 }
 
 func (re *chanResponseEmitter) CloseWithError(err error) error {
 	re.wl.Lock()
 	defer re.wl.Unlock()
 
-	return re.closeWithError(err)
+	if re.closed {
+		return errors.New("close of closed emitter")
+	}
+
+	re.closeWithError(err)
+	return nil
 }
 
-func (re *chanResponseEmitter) closeWithError(err error) error {
+func (re *chanResponseEmitter) closeWithError(err error) {
+	re.closed = true
+
 	if err == nil {
 		err = io.EOF
 	}
@@ -209,16 +208,13 @@ func (re *chanResponseEmitter) closeWithError(err error) error {
 		err = &e
 	}
 
-	re.closeOnce.Do(func() {
-		re.err = err
-		close(re.ch)
-		close(re.closed)
-	})
+	re.err = err
+	close(re.ch)
 
-	// unblock Length(), Error() and Head()
-	re.waitOnce.Do(func() {
+	// unblock Length() and Error()
+	select {
+	case <-re.wait:
+	default:
 		close(re.wait)
-	})
-
-	return nil
+	}
 }
--- a/chan.go
+++ b/chan.go
@@ -31,12 +31,8 @@ func NewChanResponsePair(req *Request) (ResponseEmitter, Response) {
 type chanStream struct {
 	req *Request
 
-	// rl is a lock for reading calls, i.e. Next.
-	rl sync.Mutex
-
 	// ch is used to send values from emitter to response.
-	// When Emit received a channel close, it sets it to nil.
-	// It is protected by rl.
+	// When Emit received a channel close, it returns the error stored in err.
 	ch chan interface{}
 
 	// wl is a lock for writing calls, i.e. Emit, Close(WithError) and SetLength.
@@ -97,10 +93,6 @@ func (r *chanResponse) Next() (interface{}, error) {
 	} else {
 		ctx = context.Background()
 	}
-
-	// to avoid races by setting r.ch to nil
-	r.rl.Lock()
-	defer r.rl.Unlock()
 
 	select {
 	case v, ok := <-r.ch:
--- a/chan.go
+++ b/chan.go
@@ -11,13 +11,11 @@ import (
 )
 
 func NewChanResponsePair(req *Request) (ResponseEmitter, Response) {
-	ch := make(chan interface{})
-	wait := make(chan struct{})
-
 	r := &chanResponse{
-		req:  req,
-		ch:   ch,
-		wait: wait,
+		req:     req,
+		ch:      make(chan interface{}),
+		waitLen: make(chan struct{}),
+		waitErr: make(chan struct{}),
 	}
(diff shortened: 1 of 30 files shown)


keks:
@Stebalien 

in one of the PRs above you said you wanted the PR bundled and the changes to go-ipfs ready. This is now done! I'm pretty pleased with the result, especially since it removes some odd stuff like that function `HandleError` or the method `RawNext` on `ResponseEmitter`.

The changes to go-ipfs are in https://github.com/ipfs/go-ipfs/pull/5035.

cli/responseemitter.go
```diff
@@ -70,26 +74,52 @@ func (re *responseEmitter) SetEncoder(enc func(io.Writer) cmds.Encoder) {
 	re.enc = enc(re.stdout)
 }
 
-func (re *responseEmitter) SetError(v interface{}, errType cmdkit.ErrorType) {
+func (re *responseEmitter) CloseWithError(err error) error {
```
whyrusleeping: It would be really great if this method didnt return an error. I can't think of any situation where the caller wants to do something with the error, or even can. I think it's likely better to just log an error and move on.

keks: The reason I let made this return an error is that `CloseWithError(error) error` is idiomatic and is used throughout the entire Go ecosystem. I don't see a reason to diverge - maybe we don't really need it in ipfs, but possible somewhere else. And now that the command's `Run` function just returns an error we don't call this so often that it's too much of a pain for the caller to do the logging themselves.

keks:
Hey @whyrusleeping judging by Stebalien's contribution activity graph it seems they are currently not avalailable, so could you maybe review this PR or delegate this? Thanks!

keks:
Also if you have questions on the structure of this bundle PR, shoot!

whyrusleeping:
@keks yeah, he's just now getting back from a vacation. I'll still want his review, but i'll take a pass here :)

chan.go
```diff
@@ -83,26 +109,33 @@ func (r *chanResponse) Next() (interface{}, error) {
...
+		if err, ok := v.(*cmdkit.Error); ok {
 			v = &err
 		}
 
 		switch val := v.(type) {
 		case *cmdkit.Error:
-			r.err = val
-			return nil, ErrRcvdError
+			// TODO keks remove logging
+			log.Error("unexpected error value:", val)
```
whyrusleeping: rm ?

chan.go
```diff
@@ -128,92 +161,94 @@ func (r *chanResponse) RawNext() (interface{}, error) {
...
-
-	length *uint64
-	err    **cmdkit.Error
-
-	emitted bool
-}
+type chanResponseEmitter chanResponse
 
 func (re *chanResponseEmitter) Emit(v interface{}) error {
+	if e, ok := v.(*cmdkit.Error); ok {
```
whyrusleeping: this is because we shouldnt be emitting errors ever anymore, right?

keks: Exactly. Should I remove this or are we not confident enough this doesn't happen? Or should I return that as an error, so it will definitely fail?

whyrusleeping: i think we should keep it, and maybe even add a check for the non-pointerized version too just in case. Then also add a comment explaining why this bit of code is here.

keks:
Hey @Stebalien, do you think you'll find some time to CR this PR and the one in go-ipfs? Would love to merge this soon

chan.go
```diff
@@ -70,6 +82,20 @@ func (r *chanResponse) Length() uint64 {
 	return r.length
 }
 
+func (re *chanResponse) Head() Head {
```
whyrusleeping: Whats this method supposed to do?

keks: It's not new, I just moved it from here: https://github.com/ipfs/go-ipfs-cmds/pull/112/files/15c1508126b5934ffae976e4a09c4353fa90956e#diff-d1319de8df19ba21c67a62fa1fde5f89L201

But I think it's needed for HEAD requests, it's even in the ResponseEmitter interface; but it looks like it's not called from anywhere in `go-ipfs-cmds/...`, so maybe we can drop it. I'd suggest doing that in a new PR though.

chan.go
```diff
@@ -57,7 +61,15 @@ func (r *chanResponse) Error() *cmdkit.Error {
 		return nil
```
Stebalien: Can we get rid of both of these nil tests? I'd rathe panic if something is wrong (and we'll panic in the second case anyways).

chan.go
```diff
@@ -70,6 +82,20 @@ func (r *chanResponse) Length() uint64 {
 	return r.length
 }
 
+func (re *chanResponse) Head() Head {
+	<-re.wait
+
+	err, ok := re.err.(*cmdkit.Error)
```
Stebalien: This will treat EOF as an error. Maybe we should just have a separate closed flag if we need that and not write EOF to the error field.

chan.go
```diff
@@ -83,26 +109,29 @@ func (r *chanResponse) Next() (interface{}, error) {
 	}
 
 	select {
+	case <-r.closed:
+		return nil, r.err
 	case v, ok := <-r.ch:
 		if !ok {
+			r.ch = nil
```
Stebalien: Are you sure this isn't racy? Also, this'll cause this method to hang forever if called again.

chan.go
```diff
@@ -83,26 +109,29 @@ func (r *chanResponse) Next() (interface{}, error) {
...
+			if r.err != nil {
+				return nil, r.err
+			}
+
 			return nil, io.EOF
 		}
 
-		if err, ok := v.(cmdkit.Error); ok {
+		if err, ok := v.(*cmdkit.Error); ok {
```
Stebalien: I'm pretty sure we just want to remove this entirely.

chan.go
```diff
@@ -128,92 +157,103 @@ func (r *chanResponse) RawNext() (interface{}, error) {
...
+	// so we want to make sure we catch situations where some code still uses the
+	// old error emitting semantics.
+	// Also errors may occur both as pointers and as plain values, so we need to
+	// check both.
+	if e, ok := v.(cmdkit.Error); ok {
+		v = &e
+	}
+	if e, ok := v.(*cmdkit.Error); ok {
+		log.Errorf("unexpected error value emitted: %s at\n%s", e.Error(), debug.Stack())
+	}
```
Stebalien: Shouldn't this call `CloseWithError` and return?

chan.go
```diff
@@ -128,92 +157,103 @@ func (r *chanResponse) RawNext() (interface{}, error) {
...
-	if re.ch == nil {
-		return nil
+func (re *chanResponseEmitter) closeWithError(err error) error {
+	if e, ok := err.(cmdkit.Error); ok {
+		err = &e
 	}
 
-	close(re.ch)
-	re.ch = nil
+	re.closeOnce.Do(func() {
```
Stebalien: We're calling this under a lock, right? We can just check if it's closed in that case. No need for a `sync.Once`.

chan.go
```diff
@@ -128,92 +157,103 @@ func (r *chanResponse) RawNext() (interface{}, error) {
...
+		re.err = err
+		close(re.ch)
+		close(re.closed)
+	})
+
+	// unblock Length(), Error() and Head()
+	re.waitOnce.Do(func() {
```
Stebalien: Same here. If we make sure to always close this under a lock (need to modify `Emit` a bit), we can just use select to see if it's still open.

chan.go
```diff
@@ -128,92 +157,103 @@ func (r *chanResponse) RawNext() (interface{}, error) {
...
 		close(re.wait)
-		re.wait = nil
-	}
+	})
+
+	re.l.Lock()
+	defer re.l.Unlock()
 
 	if re.ch == nil {
```
Stebalien: I don't think this is a valid "closed" check anymore. Probably want to check:

```go
select {
  case <-re.closed:
    return ...
  default:
}
```

cli/responseemitter.go
```diff
@@ -168,29 +210,15 @@ func (re *responseEmitter) Emit(v interface{}) error {
...
-		if t.Code == cmdkit.ErrFatal {
-			defer re.Close()
-		}
-		re.wLock.Lock()
-		defer re.wLock.Unlock()
-		re.exit = 1
-
-		return nil
-
+		log.Errorf("got an error: %v\n%s", err, debug.Stack())
```
Stebalien: Again, we should probably forward these errors to `CloseWithError` for now.

cli/responseemitter.go
```diff
@@ -168,29 +210,15 @@ func (re *responseEmitter) Emit(v interface{}) error {
...
 	case io.Reader:
 		_, err = io.Copy(re.stdout, t)
 		if err != nil {
-			re.SetError(err, cmdkit.ErrNormal)
-			err = nil
+			return err
```
Stebalien: I'm wondering if this should also call `CloseWithError`. We could also say that it's the callers responsibility to return the error from `Run`. Thoughts?

cli/responseemitter_test.go
```diff
@@ -2,10 +2,11 @@ package cli
 
 import (
 	"bytes"
+	"fmt"
 	//"io"
 	"testing"
 
-	"github.com/ipfs/go-ipfs-cmdkit"
+	//"github.com/ipfs/go-ipfs-cmdkit"
```
Stebalien: Should probably remove before final merge.

chan.go
Stebalien: Actually, we can just check `<-re.closed`.

command.go
```diff
@@ -68,24 +68,39 @@ var ErrIncorrectType = errors.New("The command returned a value with a different
 
 // Call invokes the command for the given Request
 func (c *Command) Call(req *Request, re ResponseEmitter, env Environment) {
-	// we need the named return parameter so we can change the value from defer()
-	defer re.Close()
+	var err error
+
+	defer func() {
```
Stebalien: How about putting all the logic in `Call` in a separate `call` function that turns an error. That way, we don't need this defer? The shared `err` variable makes be uncomfortable.

http/response.go
```diff
@@ -67,13 +70,29 @@ func (res *Response) RawNext() (interface{}, error) {
...
+		} else {
+			res.err = &cmdkit.Error{Message: err.Error()}
+		}
 	}
 
-	return m.Get(), err
+	return v, err
 }
 
 func (res *Response) Next() (interface{}, error) {
```
Stebalien: THis function checks if the value is an error, I assume this isn't possible given RawNext?

http/responseemitter.go
```diff
@@ -147,12 +160,28 @@ func (re *responseEmitter) Close() error {
...
+
+	re.l.Lock()
+	defer re.l.Unlock()
+
+	e, ok := err.(*cmdkit.Error)
+	if !ok {
+		e = &cmdkit.Error{Message: err.Error()}
+	}
+
+	if re.streaming || e.Code == cmdkit.ErrFatal {
```
Stebalien: Aren't all errors "fatal" now? That is, can't we just *always* use the trailer now?

writer.go
```diff
@@ -94,11 +95,10 @@ func (r *readerResponse) Next() (interface{}, error) {
 	}
 
 	switch val := v.(type) {
-	case *cmdkit.Error:
-		r.err = val
-		return nil, ErrRcvdError
 	case Single:
 		return val.Value, nil
+	case *cmdkit.Error:
```
Stebalien: Again, I assume this case is impossible.

Stebalien:
Nice! This is starting to look a lot more robust.

chan.go
keks: well the `r == nil` check obviously doesn't help because we already access `r.wait` before. But what's wrong with the `r.err == nil` check?

keks: We call `r.err.Error()` below, so `r.err` better not be nil.

chan.go
keks: You are right that this is not correct. However, there can be an error even if `re.closed` already is closed, so I don't think it's a good indicator. I think I fixed this locally and will push the changes once I'm through you first pass.

chan.go
keks: Nah, we don't emit errors anymore. Doing so is a programmer error, so we want them to know. But panicking because of this feels too strong, so I opted for logging an error.

chan.go
keks: You're right it's not a valid close check anymore, but I think your proposal is also racy: It might occur that re.closed is not yet closed but re.ch is, which would result in a write to closed channel error. Thinking this through...

chan.go
keks: Hm, yeah that would also work. I did it this way because IMHO it's more obvious what it's supposed to do than having a `select` reading from `closed` and a `default` case. This one explicitly says "do this once". Also I was under the impression that the stuff in `"sync"` usually is faster than channel ops?

chan.go
keks: I'm not sure, [here](https://github.com/ipfs/go-ipfs-cmds/pull/112#discussion_r192347895) whyrusleeping argued we should check both all the time. There's [an issue](https://github.com/ipfs/go-ipfs-cmdkit/issues/19) on fixing this for `cmdkit.Error` in general, but there was little feedback. 

Also, that line is wrong, should be without the dereference. I'll fix that.

chan.go
keks: Again, I don't think doing this with channel operations is more readable or efficient.

cli/responseemitter.go
keks: I believe we should do away with emitting errors and make sure doing so is an error.

cli/responseemitter.go
keks: Calling `CloseWithError` here would close the emitter and only print `io.Copy`'s error. I think it's better to allow the caller to specify a more intelligible error.

cli/responseemitter_test.go
keks: yup.

chan.go
Stebalien: Sorry, I was refering to both of the `r == nil` checks (in this function and the one above). Unfortunately, git hides that part of the patch.

chan.go
Stebalien: We're already checking if something is an error *before* putting it *into* the channel. If we leave this here, we'll need to leve the `cmdkit.Error` case in the switch statement below. However, IMO, we should just remove both (there shouldn't be *any* errors here).

chan.go
Stebalien: That won't cause a panic, it'll just "do the right thing" (redirect the error). As it stands, we'll end up sending the error on the value channel.

chan.go
Stebalien: For me, it's more about reducing redundant state. However, benchmarking it, they look about the same (although my benchmark may just suck).

chan.go
Stebalien: For me, every synchronization primitive is something else I have to keep track of. It's easier to keep track of one big lock (and possibly some booleans) than a bunch of `sync.Once`s.

chan.go
Stebalien: It should be if we only close the channel while holding the lock.

This is actually a good case for removing the `closeOnce` `sync.Once` (a user may think that `closeOnce` is protecting the close while really the lock is).

cli/responseemitter.go
Stebalien: We should. However, we should either *panic* in cases like this or "do the right thing". Simply logging and dropping the error on the floor is not going to end well.

cli/responseemitter.go
Stebalien: Fair enough.

chan.go
keks: Oh yes, I meant to return. Anyway, I change it to panic now. I hope we can then in the near future remove those panics because we are more confident we don't do that mistake anymore.

chan.go
keks: You're right, I got rid of closeOnce.

chan.go
keks: Unfortunately "one big lock" doesn't work in our case (because we need to `Emit` and `Next` at the same time), but I'm down to two and removed all `Once`s.

chan.go
keks: I decided to do it in this PR.

keks:
@Stebalien thank you for the thorough feedback! I tried to incorporate all of it and made some additional changes:
- remove the `Header` interface and the `Head()` methods from all `Response` types. It wasn't being used anywhere.
- remove the `RawNext` methods from all `Response` types
- remove the `HandleError` function
- unexport `WriterResponseEmitter`
- add a new global error variable `ErrClosedEmitter` that is meant to be returned by `Emit` and `Close{,WithError}`.
- work a little bit on testing. More to come in a different PR.

Right now Travis complains that `writer.go` is not gofmt'd. However, when doing `go fmt ./...`, no file is changed. I'm not sure what's going on here.

Also, in my branch of go-ipfs that has the changes for this library there is a test ([this one]) that doesn't complete currently. Investigating that now. Also my branch fell behind a bit, so I'll still need to rebase that on master -- but I first want to get that test to pass.

[this one]: https://github.com/ipfs/go-ipfs/blob/master/test/sharness/t0600-issues-and-regressions-online.sh#L49-L51

keks:
Okay, I fixed all the tests in the go-ipfs branch and rebased it is it's en par with master.

I still have no clue with Travis let's this PR fail though. It thinks that writer.go is not gofmt'd, even though it is -- at least for me calling go fmt doesn't change anything. Rebuilding doesn't change anything.
Has anyone of you stumbled over something like this in the past? My go installation is up to date.

keks:
cc @Stebalien

http/responseemitter.go
```diff
@@ -67,59 +69,57 @@ type responseEmitter struct {
 }
 
 func (re *responseEmitter) Emit(value interface{}) error {
-	ch, isChan := value.(<-chan interface{})
-	if !isChan {
-		ch, isChan = value.(chan interface{})
+	if single, ok := value.(cmds.Single); ok {
+		value = single.Value
+		defer re.closeWithError(nil)
```
Stebalien: Probably needs to go after the lock.

http/responseemitter.go
```diff
@@ -67,59 +69,57 @@ type responseEmitter struct {
...
 
 	// ignore those
 	if value == nil {
 		return nil
 	}
 
-	if _, ok := value.(cmdkit.Error); ok {
-		log.Warning("fixme: got Error not *Error: ", value)
-		value = &value
+	if err, ok := value.(cmdkit.Error); ok {
```
Stebalien: Use the debug error assertion?

writer.go
```diff
@@ -117,78 +101,88 @@ type WriterResponseEmitter struct {
...
+
+	if err == nil || err == io.EOF {
+		return re.Close()
+	}
+
+	return errors.New("provided closer does not support CloseWithError")
+}
+
+func (re *writerResponseEmitter) SetError(v interface{}, errType cmdkit.ErrorType) {
 	err := re.Emit(&cmdkit.Error{Message: fmt.Sprint(v), Code: errType})
```
Stebalien: Shouldn't we remove this function?

executor.go
```diff
@@ -105,6 +132,12 @@ func (x *executor) Execute(req *Request, re ResponseEmitter, env Environment) (e
 		}
```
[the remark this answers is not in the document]
keks: I'd still want to return the error from `errCh`, but yes the defer should also wait for it. So let's wait in both places.

executor.go
```diff
@@ -77,26 +76,54 @@ func (x *executor) Execute(req *Request, re ResponseEmitter, env Environment) (e
...
-		re.Close()
-	}()
 	defer func() {
 		// catch panics in Run (esp. from re.SetError)
 		if v := recover(); v != nil {
+			log.Errorf("panic in command handler at %s", debug.Stack())
+
 			// if they are errors
-			if e, ok := v.(error); ok {
+			if err, ok := v.(error); ok {
```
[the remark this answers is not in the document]
keks: I know it's creating a new error variable, and I use it as the argument when calling `CloseWithError`. I don't think I'm throwing it away?

writer.go
keks: We can't just remove it, because then it wouldn't satisfy the interface anymore. But I made it simply log an error.

chan.go
```diff
@@ -2,71 +2,83 @@ package cmds
...
-		return nil
-	}
+type chanResponse chanStream
 
+func (r *chanResponse) Request() *Request {
 	return r.req
 }
 
 func (r *chanResponse) Error() *cmdkit.Error {
 	<-r.wait
```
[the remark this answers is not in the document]
keks: I understand the concern about deadlocks, but returning immediately raises the question of what to return when the command is still running and we didn't see an error yet. So I went for wait-until-close, I'm currently testing this in go-ipfs to see if it makes problems.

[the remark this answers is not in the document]
keks: Okay, if we block on error until the stream is closed and change nothing else, we get a deadlock. Now I just return nil if the connection is not closed yet.

writer.go
```diff
@@ -117,78 +101,86 @@ type WriterResponseEmitter struct {
...
+	// Initially this library allowed commands to return errors by sending an
+	// error value along a stream. We removed that in favour of CloseWithError,
+	// so we want to make sure we catch situations where some code still uses the
+	// old error emitting semantics.
+	// Also errors may occur both as pointers and as plain values, so we need to
+	// check both.
+	debug.AssertNotError(v)
 
 	if _, ok := v.(Single); ok {
 		defer re.Close()
```
Stebalien: Will close without error if encoding fails.

writer.go
```diff
@@ -117,78 +101,86 @@ type WriterResponseEmitter struct {
...
+		return re.Close()
 	}
+
+	return errors.New("provided closer does not support CloseWithError")
 }
 
-func (re *WriterResponseEmitter) SetLength(length uint64) {
+func (re *writerResponseEmitter) SetError(v interface{}, errType cmdkit.ErrorType) {
+	err := &cmdkit.Error{Message: fmt.Sprint(v), Code: errType}
+	log.Errorf("error setting error %q: cannot set error with writerResponseEmitter", err)
```
Stebalien: If we can't do the right thing here, maybe we should just remove this from the interface. We really don't want the user to try calling this and then return a "success" to the user.

Stebalien: Actually, *hasn't* this been removed from the interface?

http/responseemitter.go
```diff
@@ -67,59 +69,50 @@ type responseEmitter struct {
...
-		}
-		return nil
+	re.once.Do(func() { re.preamble(value) })
+
+	re.l.Lock()
+	defer re.l.Unlock()
+	
+	if single, ok := value.(cmds.Single); ok {
+		value = single.Value
+		defer re.closeWithError(nil)
```
Stebalien: Will throw away encoding errors.

Stebalien:
FYI, if I take more than a week to get to a review, the notification has been buried in my inbox. At that point, just bug me on IRC (really, feel free to bug me on IRC repeatedly as soon as you have something for me to review).

keks:
Thank you for the review. I was traveling and will work on these over the next week. And okay, next time I'll be more stubborn :)

chan.go
```diff
@@ -85,135 +103,121 @@ func (r *chanResponse) Next() (interface{}, error) {
...
-		case *cmdkit.Error:
-			r.err = val
-			return nil, ErrRcvdError
 		case Single:
 			return val.Value, nil
 		default:
 			return v, nil
 		}
 	case <-ctx.Done():
-		close(r.done)
```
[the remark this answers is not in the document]
keks: I'm not sure. That would mean that if one call to `Next()` gets cancelled, the stream will end. Intuitively I would say that the stream should live on and the caller should be able to do another call to `Next()` later. Don't you think?

writer.go
keks: I don't think I would send the encoding error down the stream. If I try to send an unencodable value through a writer, the stream should live on and I should get an error so I can send something else.

However, we should return the `Close()` error, if there is one.

writer.go
keks: Yes, looks like I forgot deleting it. done.

http/responseemitter.go
keks: IMHO encoding errors should be passed to the caller instead of be used to close the stream. And `closeWithError` always returns `nil` so I figured we don't need it to handle the error.

keks:
Hey @Stebalien I now commented or acted on everything except the thoughts on `Single`. I'm still pondering on that one, especially whether it's possible to only set the headers that this is a single value, but still leave the closing to the caller. Otherwise making `Close()` idempotent, i.e. closing a closed channel doesn't do anything, also seems like a good idea. Giving that some thought right now.

keks:
I think the most idiomatic way is to ignore double close errors when getting an error on `Close` or `CloseWithError`, at least in places where we expect these to occur.
For this I'd introduce a new error variable `ErrEmitterClosed` and return that on double close. On the caller side I'd always check errors like
```go
err = re.Close()
if err != ErrEmitterClosed && err != nil {
  // handle error
}
```

chan.go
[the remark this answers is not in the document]
Stebalien: Ah, no, you're totally right.

Stebalien:
>  ignore this error in 

* Call
* Execute

And possibly some places in go-ipfs.

---

Warning, @whyrusleeping is grumbling about this being a massive breaking change. My plan is:

1. Finish up this PR but don't merge it:
2. *Try* patching go-ipfs.
3. If the patch is too hard to write or is unreviewable:
    1. Make the `Run` function an `interface{}`.
    2. Type switch on error/no error.
    3. Add back a `SetError` function that just calls `CloseWithError` under the hood.

http/parse.go
```diff
@@ -211,10 +212,10 @@ func parseResponse(httpRes *http.Response, req *cmds.Request) (cmds.Response, er
 			e.Message = string(mes)
 			e.Code = cmdkit.ErrNormal
 		default:
-			// handle marshalled errors
-			err := res.dec.Decode(&e)
-			if err != nil {
-				return nil, err
+			// handle errors from headers
+			errStr := httpRes.Header.Get(StreamErrHeader)
```
Stebalien: Is this correct? I think we want to set `res.initErr` and only *return* an error if we have some kind of parse error, right?

---

Also, we should probably maintain compatibility with existing servers (for now). We should probably.

1. Check the header first.
2. Check the stream next.

Speaking of which, we may want to do the inverse: Always send errors as values along with the header (just in case). Is that possible?

---

Basically, I'm worried about the case where a user calls `ipfs --api=/some/remote/server ...`.

http/response.go
```diff
@@ -74,34 +85,27 @@ func (res *Response) RawNext() (interface{}, error) {
...
-
-	return m.Get(), err
-}
-
-func (res *Response) Next() (interface{}, error) {
-	v, err := res.RawNext()
 	if err != nil {
-		return nil, err
+		if err == io.EOF {
+			res.err = err
```
Stebalien: I may be missing something, but do we need to check for trailing error headers here?

http/responseemitter.go
```diff
@@ -136,23 +129,43 @@ func (re *responseEmitter) Emit(value interface{}) error {
...
+
+func (re *responseEmitter) closeWithError(err error) error {
+	if err == io.EOF {
+		err = nil
+	}
 	if err != nil {
-		log.Debug("http.SetError err=", err)
-		panic(err)
+		// abort by sending an error trailer
+		re.w.Header().Set(StreamErrHeader, err.Error())
```
Stebalien: This is where we'd *also* send an error value if we decided to go that route. Seeing this, I'm now pretty sure we're not reading an error somewhere where we should be (I don't see us *reading* this trailer anywhere).

chan.go
```diff
@@ -85,135 +102,121 @@ func (r *chanResponse) Next() (interface{}, error) {
...
+	if e, ok := err.(cmdkit.Error); ok {
+		err = &e
+	}
+
+	re.err = err
 	close(re.ch)
-	re.ch = nil
 
-	return nil
+	// unblock Length() and Error()
```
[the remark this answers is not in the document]
keks: I was referring to both selects below that comment, but to clarify I'll just make two comments instead.

http/response.go
keks: Yes, you're totally right. Building a comprehensive test suite for cmds would be a big and important project by itself...

http/parse.go
keks: > Is this correct? I think we want to set res.initErr and only return an error if we have some kind of parse error, right?

Well, `parseResponse` is called by `Send` which is called by `Execute`, so the difference is whether you receive the error from `Send`/`Execute` or from `Next()`. IMHO returning the error from `Send` is fine, and making `Next` return the error would make the code more complex because you would have to make a noop `Response` that only returns an error and make `parseResponse` return that. I mean we can totally do that but I don't see the benefit.

> Also, we should probably maintain compatibility with existing servers (for now). We should probably...

IIUC the header is only set on connection close so I'd only check it when we receive an EOF. Does that sound reasonable to you?

> Speaking of which, we may want to do the inverse: Always send errors as values along with the header (just in case). Is that possible?

I think since the other side always checks both, it's okay to only set the header, right?

> Basically, I'm worried about the case where a user calls ipfs --api=/some/remote/server ....

Hmm, I don't get what's so special about that one. Can you elaborate?

keks: Oh I see what you mean now. Yes setting `res.initErr` seems to be the right thing to do.

cli/responseemitter.go
```diff
@@ -70,26 +71,50 @@ func (re *responseEmitter) SetEncoder(enc func(io.Writer) cmds.Encoder) {
...
+		return cmds.ErrClosingClosedEmitter
+	}
+
+	re.exit = 1 // TODO we could let err carry an exit code
 
-	err := re.Emit(&cmdkit.Error{Message: fmt.Sprint(v), Code: errType})
+	_, err = fmt.Fprintln(re.stderr, "Error:", e.Message)
 	if err != nil {
-		panic(err)
+		return err
```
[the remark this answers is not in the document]
keks: We check the errors in `Emit`, so I'd keep it in there. Also the user can specify which readers and writers to use for stdout and stderr and these don't have to be `os.Stdout` and `os.Stderr`, so it's difficult to make assumptions here.

keks:
Yup this PR is pretty big, but I have a [go-ipfs branch](https://github.com/ipfs/go-ipfs/tree/feat/cmds2/) that worked with this branch before your CR. It has fallen behind by now, but we didn't really change anything big on this side. I expect the main issue is that I'll have to rebase it to go-ipfs master.

keks:
Here is the accompanying PR: https://github.com/ipfs/go-ipfs/pull/5035

chan.go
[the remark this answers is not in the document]
Stebalien: As far as I can tell, `Error()` doesn't block on anything.

cli/responseemitter.go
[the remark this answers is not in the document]
Stebalien: Got it. Fair enough.

http/parse.go
Stebalien: > IIUC the header is only set on connection close so I'd only check it when we receive an EOF. Does that sound reasonable to you?

Yes.

> I think since the other side always checks both, it's okay to only set the header, right?

Is that *currently* true?

> Hmm, I don't get what's so special about that one. Can you elaborate?

The client version may not match the server version. If the current client doesn't always read the header or the current server version doesn't always send it, bad things may happen.

Stebalien:
> Here is the accompanying PR: ipfs/go-ipfs#5035

Nice! That's totally reviewable.

http/responseemitter.go
keks: I now check the header errors here: https://github.com/ipfs/go-ipfs-cmds/pull/112/files/#diff-c8a8dcac9ba61ed253376bf559309ae4R90

chan_test.go
```diff
@@ -78,43 +78,6 @@ func TestChanResponsePair(t *testing.T) {
 	}
 }
 
-func TestSingle1(t *testing.T) {
```
Stebalien: What was racy about this test?

cli/single_test.go
```diff
@@ -0,0 +1,54 @@
...
+	re, exitCh := NewResponseEmitter(&bufout, &buferr, cmds.Encoders["cli"], req)
+
+	wait := make(chan struct{})
+
+	go func() {
+		if err := cmds.EmitOnce(re, "test"); err != nil {
+			t.Fatal(err)
+		}
+
+		wait <- struct{}{}
```
Stebalien: Why do we need to wait here?

command.go
```diff
@@ -113,9 +114,7 @@ func (c *Command) call(req *Request, re ResponseEmitter, env Environment) error
 		} else if enc, ok := Encoders[encType]; ok {
 			re_.SetEncoder(enc(req))
 		} else {
-			// TODO or should this return an error?
-			log.Errorf("unknown encoding %q, using json", encType)
-			re_.SetEncoder(Encoders[JSON](req))
+			return fmt.Errorf("unknown encoding %q, using json", encType)
```
Stebalien: Not using json.

http/response.go
```diff
@@ -87,6 +87,12 @@ func (res *Response) Next() (interface{}, error) {
 	err := res.dec.Decode(m)
 	if err != nil {
 		if err == io.EOF {
+			// handle errors from headers
+			errStr := res.res.Header.Get(StreamErrHeader)
+			if errStr != "" {
+				return nil, &cmdkit.Error{Message: errStr}
```
Stebalien: Still needs to *set* the error.

http/responseemitter.go
Stebalien: The issue is that we're closing the response stream as if the command succeeded when, really, the command failed.

http/responseemitter.go
```diff
@@ -146,6 +143,12 @@ func (re *responseEmitter) CloseWithError(err error) error {
 	re.l.Lock()
 	defer re.l.Unlock()
 
+	if re.closed {
+		return cmds.ErrClosingClosedEmitter
+	}
+
+	re.closed = true
```
Stebalien: Shouldn't need to do this twice. We already do this in `closeWithError`.

cli/single_test.go
keks: That line is supposed to unblock the wait outside the goroutine. I added that before I read the `exitCh`, but yes we don't need that anymore.

chan_test.go
keks: I'm not 100% sure myself. The race detector complains that the [`Close` call](https://github.com/ipfs/go-ipfs-cmds/pull/112/files/1428e703db2cf9135d0c1f710fac9391f48716d1..a011832acc5d8d1bd013da9d1422f9f497989ede#diff-28e0ba9683655894160fb3fe3c0c6aceL94) and the [`Next` call](https://github.com/ipfs/go-ipfs-cmds/pull/112/files/1428e703db2cf9135d0c1f710fac9391f48716d1..a011832acc5d8d1bd013da9d1422f9f497989ede#diff-28e0ba9683655894160fb3fe3c0c6aceL101) are unsynchronized writes, but I just don't see it. The test was basically the same as `TestSingleChan`, so I removed this one.

http/parse.go
keks: Hmm, I'm afraid you are right and we'll have to send errors also as values. Damn, I hoped we would not have to do that anymore...

chan_test.go
Stebalien: Then there is a bug somewhere chanResponse.

Stebalien: Ah. I see, the test is racy.

chan_test.go
```diff
@@ -78,43 +78,6 @@ func TestChanResponsePair(t *testing.T) {
...
-		t.Fatal("error building request", err)
-	}
-	re, res := NewChanResponsePair(req)
-
-	wait := make(chan struct{})
-
-	go func() {
-		re.Emit(Single{42})
-
-		err = re.Close()
```
Stebalien: This should be `:=`

Stebalien: Fixed.

http/parse.go
```diff
@@ -190,7 +190,11 @@ func parseResponse(httpRes *http.Response, req *cmds.Request) (cmds.Response, er
 		makeDec, ok := cmds.Decoders[encType]
 		if ok {
 			res.dec = makeDec(res.rr)
+		} else {
+			log.Errorf("could not find decoder for encoding %q", encType)
```
Stebalien: Not sure if this is correct. It'll probably warn on plain text.

http/responseemitter.go
```diff
@@ -212,8 +227,10 @@ func (re *responseEmitter) doPreamble(value interface{}) {
 		} else {
 			status = http.StatusInternalServerError
 		}
+		h.Set(StreamErrHeader, err.Message)
```
Stebalien: This will set a *header* but not a *trailer*.

Stebalien: That is, it'll only work if we haven't sent a value.

chan_test.go
keks: Doh, of course! Thanks.

http/responseemitter.go
keks: Oh, apparently I deleted actually setting the trailer in `closeWithError`. I added it again, and now I always set it on non-nil errors.
Should I change that to only set the headers when we send a byte stream/`io.Reader`?

keks:
@Stebalien FYI, I rebased the branches in go-ipfs-cmds and go-ipfs and updated cmds' version in go-ipfs. This made some of the sharness tests unhappy and I'm working on getting those passing again.

Stebalien:
So, it turns out I was *wrong* about how we currently handle HTTP headers/errors. We *always* handle trailing errors using the `responseReader` helper (so we don't need to do that twice). My comment was incorrect: https://github.com/ipfs/go-ipfs-cmds/pull/112#discussion_r209087640

Given this, I believe we can switch to sending errors-in-trailers *only* for all 200 responses (don't *additionally* include the value). Thoughts? This should cover both the old client talking to a new server and a new server talking to an old client cases.

I'd like to eventually move to:

1. Sending a non-200 response on parse failure or command not found (returning the error as a plain string in the response).
2. Sending a 200 response in all other cases, returning any errors in a trailer.

But we can tackle this later.

keks:
@Stebalien done! Want to give it another round of review?

http/parse.go
```diff
@@ -218,7 +218,10 @@ func parseResponse(httpRes *http.Response, req *cmds.Request) (cmds.Response, er
 			return nil, fmt.Errorf("unknown error content type: %s", contentType)
 		default:
 			// handle errors from headers
-			e.Message = httpRes.Header.Get(StreamErrHeader)
```
Stebalien: Should we not be *returning* an error in this case?

http/responseemitter.go
```diff
@@ -112,19 +111,23 @@ func (re *responseEmitter) Emit(value interface{}) error {
...
 
-	if f, ok := re.w.(http.Flusher); ok {
-		f.Flush()
-	}
-
 	if isSingle {
-		err = re.closeWithError(err)
+		// always close singles with nil error.
+		// encoding errors go to caller.
+		err = re.closeWithError(nil)
```
Stebalien: If we return the error to the "caller", expecting them to do something, we *can't* close. This should be something like `isSingle && err != nil`. Am I missing something?

Stebalien:
You know, I think this is pretty much ready (modulo those two questions/issues). This is going to be awesome!

http/responseemitter.go
keks: It should have been `isSingle && err == nil`. If an encoding or copy error occurs, we didn't send the value so it doesn't make sense to close the emitter but we should return it. If encoding went fine, gracefully close the emitter.

http/parse.go
keks: As it stands now the error will be returned on the first call to `Next()`, but I can also make it return immediately. I believe that made the test case more complex, but maybe that's okay.

keks:
Hey @Stebalien, please CR :)

Stebalien:
@keks all ready to go! I'll merge this as soon as https://github.com/ipfs/go-ipfs/pull/5035 is ready.

PR #116, opened 2018-09-11
set the connection close header if we have a body to read

Stebalien:
We can't stream a response while reading a body *unless* we've configured the
connection to close after we're done sending a response.

This patch works around that issue by doing exactly that.

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -157,6 +157,13 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// If we have a request body, close the connection when we're done.
+	// FIXME: https://github.com/ipfs/go-ipfs/issues/5168
+	// FIXME: https://github.com/golang/go/issues/15527
+	if r.Body != http.NoBody {
+		w.Header().Set("Connection", "close")
+	}
+
 	re := NewResponseEmitter(w, r.Method, req)
 	h.root.Call(req, re, h.env)
 }


Stebalien:
This should probably have tests.

hsanjuan:
I'm wondering how this is different from disabling keep alives on the server?

I haven't read too closely, but `SetKeepAlivesEnabled(false)` seems used when managing the connection pool in the http server, so maybe manually telling clients to cut it out leaves some undesired effects?

Also, this is equivalent to asking the client to close the connection but you will need the client to comply (hopefully). Wouldn't it be better to close server-side?

Stebalien:
See: https://github.com/ipfs/go-ipfs/issues/5168#issuecomment-402806747

Basically, if we set `SetKeepAlivesEnabled(false)`, we'll cut *all* connections. We don't want to do that for the gateway and, ideally, don't want do do that for 90% of what we do anyways. With this patch, we'll ask the client to close the connection if, and only if, they've sent us the body.

Ideally, we'd only do this if they've sent us a body *and* we decide to send back a response before reading the rest of the request. However, harder to implement.

Stebalien:
@keks thoughts?

keks:
Wow this is a tricky one.

> Ideally, we'd only do this if they've sent us a body and we decide to send back a response before reading the rest of the request.

We could check this in the http response emitter `preamble` function. However, I think in there we don't know whether we read the entire body yet. Instead of reading the body directly, we could make a wrapper that just calls `req.Body.Read` and sets a flag when this returns `io.EOF`. We could then check that flag in the preamble. I'm just afraid that we might forget reading from the wrapper somewhere and instead read directly from `req.Body`... But we can overwrite that, right?

Stebalien:
> But we can overwrite that, right?

I think so... Feel free to take this over and try it.

Stebalien:
@keks does this look good to go as a stop-gap so we can fix the current issue?

keks:
@Stebalien sorry I had an exam on Wednesday. Looking into it right now.

Stebalien:
Np.

keks:
I'm still thinking whether we can come up with a good unit test for this...

http/responseemitter.go
```diff
@@ -208,6 +229,22 @@ func (re *responseEmitter) doPreamble(value interface{}) {
 		mime   string
 	)
 
+	// If we have a request body, make sure we close the body
+	// if we want to write before completing reading.
+	// FIXME: https://github.com/ipfs/go-ipfs/issues/5168
+	// FIXME: https://github.com/golang/go/issues/15527
+	if re.bodyErrChan != nil {
+		select {
+		case <-re.bodyErrChan:
```
Stebalien: I believe we can get transient errors (e.g., read timeouts). We probably do just want to check for an EOF and assume that other errors mean there may be more data. In all likelihood, any non-EOF error means the connection is in trouble anyways.

http/handler.go
```diff
@@ -109,6 +109,23 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	// If we have a request body, make sure the preamble
+	// knows that it should close the body if it wants to
+	// write before completing reading.
+	// FIXME: https://github.com/ipfs/go-ipfs/issues/5168
+	// FIXME: https://github.com/golang/go/issues/15527
+	var bodyErrChan chan error
+	if r.Body != http.NoBody {
+		bodyErrChan = make(chan error)
```
Stebalien: Needs a buffer.

keks:
Weird. I added a check that the handler actually sets the "Connection: close" header, but the client doesn't see it. However, when I look at the test coverage, the line where the header is set is green, so the handler really should set it.

http/handler.go
```diff
@@ -109,6 +110,24 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		bodyEOFChan = make(chan struct{})
+		var once sync.Once
+		bw := bodyWrapper{
+			ReadCloser: r.Body,
+			onEOF: func() {
+				once.Do(func() { close(bodyEOFChan) })
+			},
+		}
+		r.Body = bw
+	}
```
Stebalien: You may need to pre-close this channel in the else case.

keks:
@Stebalien have you seen my previous comment?

keks:
Hm? I commented on your review remark, and I think your request for change is not necessary. Do you agree? If not, why?

Stebalien:
> Hm? I commented on your review remark, and I think your request for change is not necessary. Do you agree? If not, why?

Ah, yes, you were right. That looks correct.

http/handler.go
keks: see https://github.com/ipfs/go-ipfs-cmds/pull/116#issuecomment-426586622

Stebalien:
@keks any chance we can get this moving? The tests aren't currently passing but we need to get *some* fix for this in.

PR #121, opened 2018-09-17
fix(add): refer to ipfs issue #5456

overbool:
License: MIT
Signed-off-by: Overbool <overbool.xu@gmail.com>

Fixes: https://github.com/ipfs/go-ipfs/issues/5456


overbool:
@Stebalien I'm trying to solve the https://github.com/ipfs/go-ipfs/issues/5456, but i have no idea whether my solution is correct, could you help me review my pr? Thx

Stebalien:
I think this is the correct fix but I'd like @keks to sign off on it.

Stebalien:
However, it needs a `go fmt`.

overbool:
@Stebalien @keks I had fixed it

keks:
This looks good!

@Stebalien do you think there should be a test for this edge case to prevent regressions?

Stebalien:
@keks yes. However, I like merging bug fixes, with or without tests, as long as we leave the bug open.

PR #129, opened 2018-10-26
fix encoding/decoding again

Stebalien:
This time, we completely separate error encoding and let the emitter deal with it. Basically:

1. Some transports won't even use the same channel.
2. We don't *have* an error "encoder" for protobuf.

We now *finally* return the correct content type (json) when returning an error after passing `--enc=protobuf`.

This *also*:

1. Removes GetDecoder. We need to special case this every time anyways.
2. Allows transports to specify the "default" encoding type instead of assuming JSON.
3. Ensure that we never mistake a value for an error in the HTTP transport by converting to a cmdkit.Error *before* passing errors to `doPreamble`.

Finally, this fixes a panic where we were trying to use a response emitter that we failed to construct.

### DIFF
--- a/cli/responseemitter.go
+++ b/cli/responseemitter.go
@@ -14,25 +14,21 @@ import (
 
 var _ ResponseEmitter = &responseEmitter{}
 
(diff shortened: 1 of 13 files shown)


Stebalien:
> say if this fixes all the problems

Neither can I (or anyone, as far as I know).

PR #130, opened 2018-10-26
Fix dot path parsing on Windows

djdv:
Originally fixed in: https://github.com/ipfs/go-ipfs/pull/2614
reverted in https://github.com/ipfs/go-ipfs-cmds/pull/121

Filepath was taking our IPFS friendly paths and returning system local paths.
Above is pre-patch, below is post patch
dot fix 2 0


Stebalien:
(oops)

Stebalien:
@djdv can you gx publish and make a PR against go-ipfs? I'd like to get this in before the release (will probably go to sleep soon but I'll be cutting one in the morning).

PR #134, opened 2018-11-02
MakeTypedEncoder: accept results by pointer or value

Stebalien:
Internally, we always deserialize using pointers. However, we sometimes want to deal with values *by value* instead. This makes `MakeTypedEncoder` handle both cases.

### DIFF
--- a/encoding_test.go
+++ b/encoding_test.go
@@ -36,6 +36,31 @@ func TestMakeTypedEncoder(t *testing.T) {
 	}
 }
 
+func TestMakeTypedEncoderByValue(t *testing.T) {
+	expErr := fmt.Errorf("command fooTestObj failed")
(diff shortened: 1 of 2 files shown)


encoding_test.go
```diff
@@ -36,6 +36,31 @@ func TestMakeTypedEncoder(t *testing.T) {
...
+	})
+
+	req := &Request{}
+
+	encoderFunc := f(req)
+
+	buf := new(bytes.Buffer)
+	encoder := encoderFunc(buf)
+
+	if err := encoder.Encode(&fooTestObj{true}); err != nil {
```
phritz: Shouldn't there be a test that passes fooTestObject by value?

Stebalien: So, internally we always pass to Encode by pointer (or should, at least).

PR #137, opened 2018-12-07
print a nicer error on timeout/cancel

Stebalien:
*Don't* say "context canceled". Just say "canceled".

fixes https://github.com/ipfs/go-ipfs/issues/3785

### DIFF
--- a/cli/responseemitter.go
+++ b/cli/responseemitter.go
@@ -1,13 +1,13 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
 	"sync"
 	"syscall"
 
-	"github.com/ipfs/go-ipfs-cmdkit"
 	"github.com/ipfs/go-ipfs-cmds"
 	"github.com/ipfs/go-ipfs-cmds/debug"
 )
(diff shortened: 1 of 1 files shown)


cli/responseemitter.go
```diff
@@ -63,19 +63,16 @@ func (re *responseEmitter) SetLength(l uint64) {
...
-		e = &cmdkit.Error{
-			Message: err.Error(),
-		}
+	case context.Canceled:
+		msg = "canceled"
+	case context.DeadlineExceeded:
+		msg = "timed out"
+	default:
+		msg = err.Error()
 	}
```
Stebalien: We don't need a `cmdkit.Error`, we just need the error message.

Stebalien:
This now prints:

```
Error: canceled
```

When a user hits `^C` and:

```
Error: timed out
```

When the command times out.

djdv:
I like this change but wonder if prepending "request" is necessary.
>Error:  request canceled
Error: request timed-out

vs
>Error:  canceled
Error: timed-out

Stebalien:
We can bikeshed this in a followup PR.

PR #138, opened 2018-12-24
expand on the api error

schomatis:
This would help the user figure out the problem described in https://github.com/ipfs/go-ipfs/issues/5784.

Fixes https://github.com/ipfs/go-ipfs/issues/5874.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -2,7 +2,6 @@ package http
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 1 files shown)


http/client.go
```diff
@@ -85,7 +80,7 @@ func (c *client) Execute(req *cmds.Request, re cmds.ResponseEmitter, env cmds.En
 	res, err := c.Send(req)
 	if err != nil {
 		if isConnRefused(err) {
-			err = ErrAPINotRunning
+			err = fmt.Errorf("cannot connect to the api. Is the deamon running? If this is meant to be a standalone CLI command remove the api file in `$IPFS_PATH/api`")
```
magik6k:
```suggestion
 			err = fmt.Errorf("cannot connect to the api. Is the deamon running? To run as a standalone CLI command remove the api file in `$IPFS_PATH/api`")
```

schomatis:
Fixed, thanks.

schomatis:
applied `go fmt`

Issue #113, opened 2018-08-09
Remove the "exit channel" from the CLI response emitter


Stebalien:
We should be able to encode the exit code into the error. That should clean up some complicated logic in the current CLI executor.

Issue #143, opened 2019-02-19
Wait for command to _completely_ finish before exiting.


Stebalien:
We can't just wait for the `exitCh` (from `cli.NewResponseEmitter`) to finish. We need to wait for `Run` and `PostRun` to finish. Otherwise, we don't properly close the datastore.

PR #144, opened 2019-02-19
roundup of cleanup fixes

Stebalien:
This PR includes fixes/cleanups for a bunch of issues I noticed while debugging #143.

Checkout the commit messages for details. The _important_ fix is waiting for the `Executor` to finish. before returning from `Run`. Not doing this was causing us to close the process before `Run` was actually finished.


executor.go
```diff
@@ -71,65 +69,26 @@ func (x *executor) Execute(req *Request, re ResponseEmitter, env Environment) (e
...
-				}
+				errCh <- lower.CloseWithError(cmd.PostRun[typer.Type()](res, lower))
 			}()
 		}
 	} else {
 		// not using this channel today
 		close(errCh)
 	}
 
-	defer func() {
```
Stebalien: We really shouldn't be catching panics this way. This only catches panics where the panic value is an _error_ so it's kind of arbitrary anyways.

IIRC, this was added because we had a bunch of races around `Close` and `SetError` but those have since been fixed.

examples/adder/cmd.go
```diff
@@ -184,16 +184,14 @@ var RootCmd = &cmds.Command{
 			PostRun: cmds.PostRunMap{
 				cmds.CLI: func(res cmds.Response, re cmds.ResponseEmitter) error {
 					clire := re.(cli.ResponseEmitter)
-
-					defer re.Close()
```
Stebalien: Don't need to do this explicitly anymore. It's up to the executor to close the response emitter.

cli/responseemitter.go
```diff
@@ -100,24 +67,39 @@ func (re *responseEmitter) isClosed() bool {
 }
 
 func (re *responseEmitter) Close() error {
+	return re.CloseWithError(nil)
+}
+
+func (re *responseEmitter) CloseWithError(err error) error {
```
Stebalien: IMO, it's easier to read this when _all_ the close logic is in the same function.

Stebalien:
Also fixes #113.

PR #147, opened 2019-02-27
fix sync error with go1.12 on darwin

Stebalien:
In go1.12, go switched how `file.Sync()` works on macos. Unfortunately, this also changed the "not supported" error to ENOTTY (called ENOTTY for legacy reasons) on macos.

fixes ipfs/go-ipfs#6028

### DIFF
--- a/cli/error_posix.go
+++ b/cli/error_posix.go
@@ -8,7 +8,7 @@ import (
 
 func isErrnoNotSupported(err error) bool {
 	switch err {
-	case syscall.EINVAL, syscall.ENOTSUP:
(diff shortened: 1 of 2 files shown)


cli/error_posix.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func isErrnoNotSupported(err error) bool {
 	switch err {
-	case syscall.EINVAL, syscall.ENOTSUP:
+	case syscall.EINVAL, syscall.ENOTSUP, syscall.ENOTTY:
```
Stebalien: It's called ENOTTY for historical reasons. It means "this ioctl is invalid for this file descriptor".

PR #150, opened 2019-03-01
cli: also ignore EBADF when syncing

Stebalien:
MacOS returns this when syncing stdout when piped.


Stebalien:
Will test later: https://github.com/ipfs/go-ipfs-cmds/issues/151

PR #146, opened 2019-02-25
cli: fix ignoring std{out,err} sync errors on windows

Stebalien:
fixes ipfs/go-ipfs#6021

PR #153, opened 2019-03-04
cli: don't sync output to NUL on Windows

djdv:
I know the documentation seems backwards, but it's not.
fixes ipfs/go-ipfs#6039
followup to https://github.com/ipfs/go-ipfs-cmds/pull/146


Stebalien:
Thanks!

PR #155, opened 2019-03-21
typed encoder: improve pointer reflection

Stebalien:
Allow value -> ptr in addition to ptr -> value conversion.


Kubuxu:
The comment I raised in the PR was about it not giving me any errors but this works as well.

Stebalien:
It may have sent the errors over the API, I'm not sure.

Stebalien:
@Kubuxu is that an approval?

PR #154, opened 2019-03-21
support url paths as files.WebFile

jmank88:
This PR updates the `cli` parser to create `files.WebFile`s for URL paths. ~It also updates the `go-ipfs-files` dependency to `v0.0.2` (to support `nocopy` from https://github.com/ipfs/go-ipfs-files/pull/13). This version is not strictly required to compile or use URLs alone (what really matter is which version `go-ipfs` uses), but simplifies things by coupling `nocopy` support from the start.~

Towards https://github.com/ipfs/go-ipfs/issues/6065
~Draft until `go-ipfs-files` `v0.0.2` is released.~

PR #158, opened 2019-03-22
set WebFile fpath to URL base

jmank88:
This PR sets the `fpath` for `WebFile`s to the 'base' of the URL, which is the base of the *path* if non-empty, otherwise the host.
Follow up from #154


jmank88:
Now `ipfs add "http://test/test" "http://different/test" "http://another/test" "http://more/test?param=value"` produces:
```
test
test (1)
test (2)
test (3)
```

PR #163, opened 2019-04-30
http: use the request context

Stebalien:
CloseNotifier has been deprecated for a while.

Also, ditch the "environment" context. We don't actually _need_ this.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -8,12 +8,8 @@ type Executor interface {
 	Execute(req *Request, re ResponseEmitter, env Environment) error
 }
 
-// Environment is the environment passed to commands. The only required method
-// is Context.
-type Environment interface {
-	// Context returns the environment's context.
-	Context() context.Context
-}
+// Environment is the environment passed to commands.
+type Environment interface{}
 
 // MakeEnvironment takes a context and the request to construct the environment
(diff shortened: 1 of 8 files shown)


executor.go
```diff
@@ -8,12 +8,8 @@ type Executor interface {
...
 }
 
-// Environment is the environment passed to commands. The only required method
-// is Context.
-type Environment interface {
-	// Context returns the environment's context.
-	Context() context.Context
-}
+// Environment is the environment passed to commands.
+type Environment interface{}
```
Stebalien: (doesn't break anything in either go-filecoin or go-ipfs)

djdv: Should `MakeEnvironment` be removed with this?

Stebalien: We still need the environment (we type assert it to something more useful).

Really, we should consider stashing everything we'd _usually_ put in the environment in the _context_ (which is, IMO, where it should go) but we can handle that later.

http/parse.go
```diff
@@ -1,7 +1,8 @@
 package http
 
 import (
-	"context"
+	"crypto/rand"
```
magik6k: May want to use 'math/rand', it's 10-30x faster

PR #164, opened 2019-05-11
merge in go-ipfs-cmdkit

Stebalien:
This PR merges go-ipfs-cmdkit back into go-ipfs-cmds. That package was extracted as a common dependency for the legacy commands library and the new one but the split is no longer necessary (and is kind of a pain).

cmdkit PR: https://github.com/ipfs/go-ipfs-cmdkit/pull/34

Specifically, this just:

1. Imports all relevant commits from go-ipfs-cmdkit.
2. Renames cmdkit -> cmds.
3. Removes some dead code.


Stebalien:
Ya know, nobody's going to object to this and it "just works". :shipit:.

PR #171, opened 2019-06-13
fallback executor support

Stebalien:
This adds fallback support to the HTTP executor. That way, we can try to run a command locally if the daemon is offline.

Note: This is an API breaking change which I will bubble up to both go-ipfs and filecoin. Specifically, `http.NewClient` now returns an `Executor`, not an `http.Client`. The correct way to use this client was to cast it to an executor.

### DIFF
--- a/examples/adder/remote/client/main.go
+++ b/examples/adder/remote/client/main.go
@@ -19,16 +19,10 @@ func main() {
 		panic(err)
 	}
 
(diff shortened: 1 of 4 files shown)


Stebalien:
For:

* https://github.com/ipfs/go-ipfs/issues/5784
* https://github.com/ipfs/go-ipfs/issues/6435 (so we can get rid of the repo fsck command entirely)

PR #173, opened 2019-08-22
http: cleanup http related errors

Stebalien:
We don't need to return information about the actual HTTP error. That's usually
not something the user cares about (and is usually implied from the command).

This way, we get an actual context canceled error when the context is canceled.


Stebalien:
Relates to https://discuss.ipfs.io/t/the-get-commands-is-http-request/6136.

PR #184, opened 2020-03-16
Special-case accepting explicitly supplied named pipes

ribasushi:
This is properly tested in go-ipfs ( can't open PR before bubbling this as a release )


cli/parse.go
```diff
@@ -510,6 +510,15 @@ func appendFile(fpath string, argDef *cmds.Argument, recursive, hidden bool) (fi
...
 		}
+	} else if (stat.Mode() & os.ModeNamedPipe) != 0 {
+		// Special case pipes that are provided directly on the command line
+		// We do this here instead of go-ipfs-files, as we need to differentiate between
+		// recursive(unsupported) and direct(supported) mode
+		file, err := os.Open(fpath)
+		if err != nil {
+			return nil, err
+		}
+		return files.NewReaderPathFile(fpath, file, stat)
```
Stebalien: The path & stat should be nil.

* The path should only refer to files that can _actually_ be read. The filestore/urlstore features will use it.
* The stat will be used to determine the size of the file. It'll give us _very_ incorrect information in this case.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # go-ipfs changelog
 
+- Accept named fifos ( pipes ) when supplied directly as an argument, continues to reject them when encountered recursively
```
Stebalien: Ah, let's just delete this file. This was the go-ipfs changelog and somehow snuck into this repo when the commands lib package was extracted.

cli/parse.go
ribasushi: `path` is a string, so I went with `""` ( thought of using `newMessageReader` but proably incorrect in this case ). There were zero changes needed in the go-ipfs test suite ( aside from https://github.com/ipfs/go-ipfs/pull/6998 obviously ), so this seems like the right thing to do.

PR #190, opened 2020-04-02
http: configurable allowed request methods for the API.

hsanjuan:
Effectively enforce that the IPFS API is an RPC API that only handles POST requests.

### DIFF
--- a/http/config.go
+++ b/http/config.go
@@ -22,6 +22,12 @@ type ServerConfig struct {
 	// Headers is an optional map of headers that is written out.
 	Headers map[string][]string
 
+	// HandledMethods set which methods will be handled for the HTTP
+	// requests. Other methods will return 405. This is different from CORS
+	// AllowedMethods (the API may handle GET and POST, but only allow GETs
+	// for CORS-enabled requests via AllowedMethods).
+	HandledMethods []string
+
 	// corsOpts is a set of options for CORS headers.
 	corsOpts *cors.Options
 
@@ -32,6 +38,7 @@ type ServerConfig struct {
 func NewServerConfig() *ServerConfig {
 	cfg := new(ServerConfig)
 	cfg.corsOpts = new(cors.Options)
+	cfg.HandledMethods = []string{http.MethodPost}
 	return cfg
 }
 
@@ -140,5 +147,18 @@ func allowReferer(r *http.Request, cfg *ServerConfig) bool {
 		}
 	}
 
+	return false
+}
+
+// handleRequestMethod returns true if the request method is among
+// HandledMethods.
+func handleRequestMethod(r *http.Request, cfg *ServerConfig) bool {
+	// For very small slices as these, this should be faster than
+	// a map lookup.
+	for _, m := range cfg.HandledMethods {
+		if r.Method == m {
+			return true
+		}
+	}
 	return false
 }
--- a/http/errors_test.go
+++ b/http/errors_test.go
@@ -9,7 +9,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/ipfs/go-ipfs-cmds"
+	cmds "github.com/ipfs/go-ipfs-cmds"
 )
 
 func TestErrors(t *testing.T) {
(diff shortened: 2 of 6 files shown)


hsanjuan:
I need to fix tests... tomorrow...

hsanjuan:
> I don't think we can do this here because we still want to support GET for read-only requests on the read-only API. Let's do this in go-ipfs itself.

I am going to argue against this. This causes ambiguity between two APIs which are the same. Tools written to work on the gateway endpoint may not work on the API endpoint even though they are supposed to be using the same API.

More importantly, it does not fully fix our current problem. The fact that this is a Read-Only API does not mean that I wish to give anyone capabilities to, for example, `pin ls`. Apart from privacy, it is a DDoS vector. More worryingly, afaik you can use the Read only API to `cat`, which efectivelly means making the node retrieve any content whatsoever and write it to disk.

Reconsider please?

hsanjuan:
> More worryingly, afaik you can use the Read-only API to cat, which efectivelly means making the node retrieve any content whatsoever and write it to disk.

After sleeping, I realize that this runs on the gateway endpoint anyway so it should not matter as that is effectively cat'ing too. Still we have seen people complain that it is not obvious that a read-only API is running side-by-side to the gateway, and thus may lead to situations where people forget to secure these paths like they would secure /ipns /ipfs when putting behind nginx etc. The other issues still stand I think... at least in some setups.

Stebalien:
We really need to:

1. Flesh out the gateway documentation.
2. Have a flag to disable this.

Stebalien:
Discussed async: make make this an option.

http/config.go
```diff
@@ -142,3 +148,16 @@ func allowReferer(r *http.Request, cfg *ServerConfig) bool {
 
 	return false
 }
+
+// handleRequestMethod returns true if the request method is among
+// HandledMethods.
+func handleRequestMethod(r *http.Request, cfg *ServerConfig) bool {
+	// For very small slices as these, this should be faster than
+	// a map lookup.
```
hsanjuan, line 156: Someone did the numbers https://boltandnuts.wordpress.com/2017/11/20/go-slice-vs-maps/

hsanjuan:
> The fact that this is a Read-Only API does not mean that I wish to give anyone capabilities to, for example, `pin ls`.

For anyone following along, I misread: it's `ls` capabilities, not pin ls. The functionality offered by the read-only simply amounts to what the gateway offers already but with a more specialized tooling. Thus we can keep it.

http/config.go
Stebalien: Be careful. Go is constantly changing.

But this isn't performance critical anyways.

http/config.go
```diff
@@ -22,6 +22,12 @@ type ServerConfig struct {
 	// Headers is an optional map of headers that is written out.
 	Headers map[string][]string
 
+	// HandledMethods set which methods will be handled for the HTTP
+	// requests. Other methods will return 405. This is different from CORS
+	// AllowedMethods (the API may handle GET and POST, but only allow GETs
+	// for CORS-enabled requests via AllowedMethods).
+	HandledMethods []string
```
Stebalien, line 29: Should default to POST in `NewServerConfig`.

hsanjuan:
Bah let me add a test

hsanjuan:
I have squashed and improved the commit msg.

hsanjuan:
@Stebalien technically you should review new changes to `_test` files. Merge when ready. I have a go-ipfs PR ready for when this can bubble, can send tomorrow.

http/config.go
Stebalien: Result on go 1.13 mapping string -> bool:

* 1 item - array is faster.
* 2 items - same time.
* 3 items - map is faster.

http/config.go
```diff
@@ -22,6 +22,12 @@ type ServerConfig struct {
 	// Headers is an optional map of headers that is written out.
 	Headers map[string][]string
 
+	// HandledMethods set which methods will be handled for the HTTP
+	// requests. Other methods will return 405. This is different from CORS
+	// AllowedMethods (the API may handle GET and POST, but only allow GETs
+	// for CORS-enabled requests via AllowedMethods).
```
Stebalien, line 28: Hm. Actually, what is this doing?

Stebalien: nvm

http/config.go
hsanjuan: wow

PR #191, opened 2020-04-04
change HandledMethods to AllowGet and cleanup method handling

Stebalien:
Allowing methods isn't as simple as just allowing/disallowing them because different methods do different things.

* HEAD: should be allowed everywhere GET is allowed.
* OPTIONS:
  * When a CORS request is made, this will be handled by the CORS library.
  * Otherewise, we need to return the allowed methods.
* POST: always allowed.
* Everything else: always denied.

Changing HandledMethods to a simple AllowGet makes it easier to "do the right thing".

### DIFF
--- a/http/config.go
+++ b/http/config.go
@@ -22,11 +22,13 @@ type ServerConfig struct {
 	// Headers is an optional map of headers that is written out.
 	Headers map[string][]string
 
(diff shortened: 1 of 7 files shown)


hsanjuan:
I'll bubble to go-ipfs

PR #193, opened 2020-04-06
Fix: disallow POST without Origin nor Referer from specific user agents

hsanjuan:
Addresses browsers being able to POST without control due to things like
https://bugzilla.mozilla.org/show_bug.cgi?id=429594

### DIFF
--- a/http/config.go
+++ b/http/config.go
@@ -3,6 +3,7 @@ package http
 import (
 	"net/http"
 	"net/url"
+	"strings"
 	"sync"
 
 	cors "github.com/rs/cors"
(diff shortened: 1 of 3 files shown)


hsanjuan:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent/Firefox

The problem affects Firefox, but what about:

* Focus/Klar for Android: Different UA only in old versions, now it's Firefox it seems.
* Focus for FireTV and Echo Show: seems to use `Focus` user agent still.
* `FxiOS` (Focus and Firefox for iOS)
* Fennec/Seamonkey etc: should include `Firefox` in the UA.

hsanjuan:
But all of those would not have IPFS API running on a local port so it should be fine...

hsanjuan:
Apparently Android at least can run go-ipfs on 127.0.0.1:5001, so I have included Focus, Klar, FxiOS in the list. It should not affect them if they are setting the headers right anyways.

But that said, we could just run this check on anything `^Mozilla` and catch all browsers ?

PR #196, opened 2020-04-29
Staticcheck fixes

hsanjuan:
This fixes all staticcheck warnings for this library.

### DIFF
--- a/cli/cmd_suggestion.go
+++ b/cli/cmd_suggestion.go
@@ -78,12 +78,18 @@ func suggestUnknownCmd(args []string, root *cmds.Command) []string {
 func printSuggestions(inputs []string, root *cmds.Command) (err error) {
 
 	suggestions := suggestUnknownCmd(inputs, root)
+
 	if len(suggestions) > 1 {
-		err = fmt.Errorf("unknown Command \"%s\"\n\nDid you mean any of these?\n\n\t%s", inputs[0], strings.Join(suggestions, "\n\t"))
+		//lint:ignore ST1005 user facing error
+		err = fmt.Errorf("Unknown Command \"%s\"\n\nDid you mean any of these?\n\n\t%s", inputs[0], strings.Join(suggestions, "\n\t"))
+
 	} else if len(suggestions) > 0 {
-		err = fmt.Errorf("unknown Command \"%s\"\n\nDid you mean this?\n\n\t%s", inputs[0], suggestions[0])
+		//lint:ignore ST1005 user facing error
+		err = fmt.Errorf("Unknown Command \"%s\"\n\nDid you mean this?\n\n\t%s", inputs[0], suggestions[0])
+
 	} else {
-		err = fmt.Errorf("unknown Command \"%s\"", inputs[0])
+		//lint:ignore ST1005 user facing error
+		err = fmt.Errorf("Unknown Command \"%s\"\n", inputs[0])
 	}
 	return
 }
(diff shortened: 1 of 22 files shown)


http/handler.go
```diff
@@ -17,11 +16,12 @@ import (
 var log = logging.Logger("cmds/http")
 
 var (
-	ErrNotFound           = errors.New("404 page not found")
-	errApiVersionMismatch = errors.New("api version mismatch")
```
hsanjuan, line 21: unused

http/responseemitter.go
```diff
@@ -75,7 +77,7 @@ type responseEmitter struct {
 
 	l      sync.Mutex
 	length uint64
-	err    *cmds.Error
+	// err    *cmds.Error
```
hsanjuan, line 80: Field unused, maybe should remove? Or leave commented?

http/responseemitter.go
```diff
@@ -8,14 +8,14 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/ipfs/go-ipfs-cmds"
+	cmds "github.com/ipfs/go-ipfs-cmds"
 )
 
 var (
-	HeadRequest = fmt.Errorf("HEAD request")
```
hsanjuan, line 15: I think this is the only exported thing that is changing. I don't think it is used anywhere in our codebases. Seemed ok to remove.

cli/cmd_suggestion.go
```diff
@@ -79,11 +79,11 @@ func printSuggestions(inputs []string, root *cmds.Command) (err error) {
 
 	suggestions := suggestUnknownCmd(inputs, root)
 	if len(suggestions) > 1 {
-		err = fmt.Errorf("Unknown Command \"%s\"\n\nDid you mean any of these?\n\n\t%s", inputs[0], strings.Join(suggestions, "\n\t"))
+		err = fmt.Errorf("unknown Command \"%s\"\n\nDid you mean any of these?\n\n\t%s", inputs[0], strings.Join(suggestions, "\n\t"))
 	} else if len(suggestions) > 0 {
-		err = fmt.Errorf("Unknown Command \"%s\"\n\nDid you mean this?\n\n\t%s", inputs[0], suggestions[0])
+		err = fmt.Errorf("unknown Command \"%s\"\n\nDid you mean this?\n\n\t%s", inputs[0], suggestions[0])
 	} else {
-		err = fmt.Errorf("Unknown Command \"%s\"\n", inputs[0])
```
hsanjuan, line 86: This is going to alter some formatting in the output of go-ipfs. However I am not sure we should do formatting like this. Ideas?

http/responseemitter.go
Stebalien: Remove. We used to have a `SetError` function but switched to simply returning errors.

cli/cmd_suggestion.go
Stebalien: I'd leave these as-is and add a `//nolint` comment. These are really just user-facing error messages.

command.go
```diff
@@ -93,13 +94,13 @@ type Command struct {
 
 var (
 	// ErrNotCallable signals a command that cannot be called.
-	ErrNotCallable = ClientError("This command can't be called directly. Try one of its subcommands.")
+	ErrNotCallable = ClientError("this command can't be called directly. Try one of its subcommands.")
```
Stebalien, line 97: I wouldn't fix this unless we get rid of all caps: "this command can't be called directly; try one of its subcommands"

hsanjuan:
I made the changes suggested. Merging as it was pre-approved and this should be low risk.

PR #197, opened 2020-04-29
doc: update docs for interface changes

Stebalien:


Stebalien:
I only switched it because I've added the `CloseWithError` function.

PR #198, opened 2020-07-30
refactor: move external command checks into commands lib

aschmahmann:


command.go
```diff
@@ -90,6 +90,39 @@ type Command struct {
...
+	// NoLocal denotes that a command cannot be executed in a local environment
+	NoLocal bool
+
+	// Extra contains a set of other command-specific parameters
+	Extra *Extra
+}
+
+// Extra is a set of tag information for a command
+type Extra struct {
+	m map[interface{}]interface{}
```
Stebalien, line 106: I'd consider just using `type Extra map[interface{}]interface{}`. That'll let users configure this using the map syntax (`Extra{SomeKnownKey: someValue}`).

http/parse.go
```diff
@@ -28,12 +28,19 @@ func parseRequest(r *http.Request, root *cmds.Command) (*cmds.Request, error) {
...
 
-	cmd, err := root.Get(getPath)
+	cmdPath, err := root.Resolve(getPath)
 	if err != nil {
 		// 404 if there is no command at that path
 		return nil, ErrNotFound
 	}
 
+	for _, c := range cmdPath {
+		if c != nil && c.NoRemote {
```
Stebalien, line 38: I don't think we need the nil check here. `getPath` is the path up to but not including the last element.

We need the nil check below because the last element may be either a value or a subcommand. However, none of the elements of `cmdPath` should be nil (`Resolve` returns an error if it fails to find a subcommand).

aschmahmann: > Resolve returns an error if it fails to find a subcommand

Yep, I'll remove this.

command.go
aschmahmann: I was trying to stay away from exposing `SomeKnownKey` since thus far in our code base I've seen functions controlling the getting/setting of the `type mymarker struct{}` map entries (e.g. for the libp2p routing events).

We could take your suggestion, although we'd probably need to document the footgun of using the map directly instead of the Get method (i.e. the map may be nil).

Seem worth it to you? If so lmk and I'll make the change.

Stebalien: Eh, it's fine either way.

aschmahmann: k, going to leave it as is and merge + bubble

PR #199, opened 2020-08-01
Fix build on Plan 9

fhs:


willscott:
the posix/windows implementations include EINVAL and ENOTSUP style errors in the check.
plan9 also at least defines these (`syscall.EINVAL` and `syscall.EPLAN9`).  Is there a reason those shouldn't bubble up to returning true in `isErrnoNotSupported`?

fhs:
> the posix/windows implementations include EINVAL and ENOTSUP style errors in the check.
> plan9 also at least defines these (`syscall.EINVAL` and `syscall.EPLAN9`). Is there a reason those shouldn't bubble up to returning true in `isErrnoNotSupported`?

This function seems to be only called with an unwrapped `os.PathError` error returned by `File.Sync`. `File.Sync` can return `os.ErrInvalid` for nil receiver, but it's not wrapped in `os.PathError`. I don't think it ever returns `syscall.EINVAL` or `syscall.EPLAN9`. I can return true for those errors if the function is meant for general use, not just to check `File.Sync` errors.

willscott:
>  I can return true for those errors if the function is meant for general use, not just to check `File.Sync` errors.

It seems like it shouldn't hurt, and hopefully reduces the potential for surprise in the future if it stays consistent with other platforms it case it gets used for something else

fhs:
I've added `syscall.EINVAL` and `syscall.EPLAN9`. `os.ErrInvalid` is not being checked on unix or windows, so leaving that out.

PR #202, opened 2020-10-22
fix: hide subcommand from short help when no subcommands

olizilla:
ShortHelp always shows the subcommand help prompt even if there are no subcommands. In `ipfs@0.7`

```console
$ ipfs version
ipfs version 0.7.0

$ipfs ls -h
USAGE
  ipfs ls <ipfs-path>... - List directory contents for Unix filesystem objects.

  ipfs ls [--headers | -v] [--resolve-type=false] [--size=false] [--stream | -s] [--] <ipfs-path>...

  Displays the contents of an IPFS or IPNS object(s) at the given path, with
  the following format:
  
    <link base58 hash> <link size in bytes> <link name>
  
  The JSON output contains type information.

  'ipfs ls <subcmd> --help'
```

The last 2 lines of the short help are supposed to let the user know that `--help` will get them more help about *this* command. There should be no invitation to check subcommands if there are no subcommands for this command.

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>

### DIFF
--- a/cli/helptext_test.go
+++ b/cli/helptext_test.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"bytes"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


cli/helptext_test.go
```diff
@@ -48,3 +49,47 @@ func TestSynopsisGenerator(t *testing.T) {
 		t.Fatal("Synopsis should contain options finalizer")
 	}
 }
+
+func TestShortHelp(t *testing.T) {
```
aschmahmann, line 53: Could you also add a test that checks on subcommands? The two things I'd like to watch for are:
1) I'd like some signal that these tests are still useful even if someone changes "For more information about each command" to be rephrased slightly. This could also be aided by checking a constant.
2) Making sure that subcommands are printed

cli/helptext.go
```diff
@@ -110,9 +110,11 @@ const shortHelpFormat = `USAGE
 {{end}}{{if .Subcommands}}
 SUBCOMMANDS
 {{.Subcommands}}
-{{end}}{{if .MoreHelp}}
 {{.Indent}}For more information about each command, use:
 {{.Indent}}'{{.Path}} <subcmd> --help'
+{{end}}{{if .MoreHelp}}
+{{.Indent}}for more information about this command, use:
```
aschmahmann, line 116: lowercase f here intentional?

aschmahmann (commented):
Left a few tiny suggestions/comments but LGTM. Thanks 😄.

Note: Does this effect our documentation generation at all or does it not utilize this codepath?

PR #203, opened 2020-11-25
feat: support strings option over HTTP API

aschmahmann:
StringsOption was not properly supported over HTTP API before only over the CLI. This resolves that issue.

Note: Still needs tests as this was previously thought to be supported but was not.


http/parse.go
```diff
@@ -61,22 +61,28 @@ func parseRequest(r *http.Request, root *cmds.Command) (*cmds.Request, error) {
...
 			name := optDef.Names()[0]
 			if k != name {
-				opts[name] = v
-				delete(opts, k)
+				iopts[name] = v
+				delete(iopts, k)
+			}
+
+			if optDef.Type() != cmds.Strings && len(v) > 0 {
+				iopts[name] = v[0]
```
aschmahmann, line 79: Should we set `iopts[name]` to empty if `len(v) == 0`?

Stebalien: We should probably fail if we get something we don't expect. That includes an empty value (should be impossible) and multiple values where exactly one is expected.

http/parse.go
```diff
@@ -61,22 +61,28 @@ func parseRequest(r *http.Request, root *cmds.Command) (*cmds.Request, error) {
...
 	}
 	for k, v := range opts {
+		iopts[k] = v
 		if optDef, ok := optDefs[k]; ok {
 			name := optDef.Names()[0]
 			if k != name {
-				opts[name] = v
-				delete(opts, k)
+				iopts[name] = v
+				delete(iopts, k)
```
Stebalien, line 75: I'd clean this up a bit. Given this logic, there's no need to set the value in `iopts`, then delete it. We can just set it once.

http/parse.go
```diff
@@ -162,17 +168,16 @@ func parseRequest(r *http.Request, root *cmds.Command) (*cmds.Request, error) {
 	return req, err
 }
 
-func parseOptions(r *http.Request) (map[string]interface{}, []string) {
-	opts := make(map[string]interface{})
+func parseOptions(r *http.Request) (map[string][]string, []string) {
```
Stebalien, line 171: I'd consider removing this helper. Up to you but it's looking like that might actually simplify the code at this point.

request.go
```diff
@@ -118,21 +118,28 @@ func checkAndConvertOptions(root *Command, opts OptMap, path []string) (OptMap,
...
+				if _, ok := v.([]string); !ok {
+					return options, fmt.Errorf("option %q should be type %q, but got type %q",
+						k, opt.Type().String(), kind.String())
 				}
-				options[k] = val
-
 			} else {
-				return options, fmt.Errorf("option %q should be type %q, but got type %q",
-					k, opt.Type().String(), kind.String())
+				if str, ok := v.(string); ok {
```
Stebalien, line 127: We could invert this check to drop the else clause (and use an early return).

Stebalien (approved):
Some small nits but otherwise LGTM.

http/client.go
```diff
@@ -230,6 +230,15 @@ func getQuery(req *cmds.Request) (string, error) {
...
+
+		optArr, ok := v.([]string)
+		if ok {
+			for _, o := range optArr {
+				query.Add(k, o)
+			}
+			continue
+		}
+
 		str := fmt.Sprintf("%v", v)
```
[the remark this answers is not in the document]
aschmahmann: done

http/parse.go
aschmahmann: done

http/parse.go
aschmahmann: done

request.go
aschmahmann: done (although wasn't sure what to call this commit so just used "refactor: cleanup code"

http/parse.go
aschmahmann: done. although it turns out `len(v)==0` isn't a thing as `URL.Query()` just returns `[]string{""}` if there are no values.

PR #204, opened 2021-01-15
StringsOption with delimiter

aschmahmann:
Added a variant of StringsOption that supports getting delimited lists on the CLI for conciseness.

Also, made `WithDefault` panic when passing in the wrong type.

Note: Delimited only works on the CLI. This means we should probably return an error if the HTTP API tries to pass a string with a delimiter in it, but that's not done yet. Callers should handle bad strings so this doesn't seem particularly problematic.

### DIFF
--- a/cli/parse_test.go
+++ b/cli/parse_test.go
@@ -84,6 +84,7 @@ func TestOptionParsing(t *testing.T) {
 			cmds.StringOption("flag", "alias", "multiple long"),
 			cmds.BoolOption("bool", "b", "a bool"),
 			cmds.StringsOption("strings", "r", "strings array"),
+			cmds.DelimitedStringsOption(",", "delimstrings", "d", "comma delimited string array"),
 		},
 		Subcommands: map[string]*cmds.Command{
 			"test": &cmds.Command{},
@@ -154,6 +155,13 @@ func TestOptionParsing(t *testing.T) {
 	test("-b --string foo test bar", kvs{"bool": true, "string": "foo"}, words{"bar"})
 	test("-b=false --string bar", kvs{"bool": false, "string": "bar"}, words{})
 	test("--strings a --strings b", kvs{"strings": []string{"a", "b"}}, words{})
+
+	test("--delimstrings a,b", kvs{"delimstrings": []string{"a", "b"}}, words{})
+	test("--delimstrings=a,b", kvs{"delimstrings": []string{"a", "b"}}, words{})
+	test("-d a,b", kvs{"delimstrings": []string{"a", "b"}}, words{})
+	test("-d=a,b", kvs{"delimstrings": []string{"a", "b"}}, words{})
+	test("-d=a,b -d c --delimstrings d", kvs{"delimstrings": []string{"a", "b", "c", "d"}}, words{})
+
 	testFail("foo test")
 	test("defaults", kvs{"opt": "def"}, words{})
 	test("defaults -o foo", kvs{"opt": "foo"}, words{})
--- a/cli/parse_test.go
+++ b/cli/parse_test.go
@@ -77,6 +77,24 @@ func TestSameWords(t *testing.T) {
 	test(f, f, true)
 }
 
+func testOptionHelper(t *testing.T, cmd *cmds.Command, args string, expectedOpts kvs, expectedWords words, expectErr bool) {
+	req := &cmds.Request{}
+	err := parse(req, strings.Split(args, " "), cmd)
+	if err == nil {
+		err = req.FillDefaults()
+	}
+	if expectErr {
+		if err == nil {
+			t.Errorf("Command line '%v' parsing should have failed", args)
+		}
+	} else if err != nil {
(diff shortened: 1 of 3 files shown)


option.go
```diff
@@ -184,6 +196,44 @@ func FloatOption(names ...string) Option {
...
+	return DelimitedStringsOption("", names...)
+}
+
+// DelimitedStringsOption like StringsOption is a command option that can handle a slice of strings.
+// However, DelimitedStringsOption will automatically break up the associated CLI inputs based on the delimiter.
+// For example, instead of passing `command --option=val1 --option=val2` you can pass `command --option=val1,val2` or
+// even `command --option=val1,val2 --option=val3,val4`.
+//
+// A delimiter of "" means that no delimiter is used
+func DelimitedStringsOption(delimiter string, names ...string) Option {
```
aschmahmann, line 211: Went with two different names here because just adding another string option called delimiter in front will not cause a compile error for any users of the current non-delimiter StringsOption which seems useful (at least while drafting the PR if not further).

Options I can see so far include:
- Two options as the PR has now
- Just one option, but making sure to fix all the cases in this library and go-ipfs (seems error prone, but doable)
- Use one option, but instead of a string delimiter pass in a non-convertable type or option so that we get a compile error

option.go
```diff
@@ -184,6 +196,44 @@ func FloatOption(names ...string) Option {
...
-	return NewOption(Strings, names...)
+	return DelimitedStringsOption("", names...)
+}
+
+// DelimitedStringsOption like StringsOption is a command option that can handle a slice of strings.
+// However, DelimitedStringsOption will automatically break up the associated CLI inputs based on the delimiter.
+// For example, instead of passing `command --option=val1 --option=val2` you can pass `command --option=val1,val2` or
+// even `command --option=val1,val2 --option=val3,val4`.
+//
+// A delimiter of "" means that no delimiter is used
```
aschmahmann, line 210: This is as opposed to strings.Split where an empty delimiter means to split on every character. We could add this behavior, but I'm not currently seeing the utility.

option.go
```diff
@@ -149,6 +149,18 @@ func NewOption(kind reflect.Kind, names ...string) Option {
 }
 
 func (o *option) WithDefault(v interface{}) Option {
+	if v == nil {
+		panic(fmt.Errorf("cannot use nil as a default"))
+	}
```
aschmahmann, line 154: I'm not seeing any problem with this ATM, it seems like so far all of our types are concrete and even for the pointer types we do have (i.e. []string) as long as people pass in `var emptyStingrArray []string` instead of just the untyped `nil` it will be fine.

cli/parse_test.go
```diff
@@ -77,13 +77,32 @@ func TestSameWords(t *testing.T) {
 	test(f, f, true)
 }
 
+func testOptionHelper(t *testing.T, cmd *cmds.Command, args string, expectedOpts kvs, expectedWords words, expectErr bool) {
```
aschmahmann, line 80: this was just extracted from the function below

option.go
```diff
@@ -149,6 +149,18 @@ func NewOption(kind reflect.Kind, names ...string) Option {
...
+
+	// if type of value does not match the option type
+	if vKind, oKind := reflect.TypeOf(v).Kind(), o.Type(); vKind != oKind {
+		// if the reason they do not match is not because of Slice vs Array equivalence
+		// Note: Figuring out if the type of Slice/Array matches is not done in this function
+		if !((vKind == reflect.Array || vKind == reflect.Slice) && (oKind == reflect.Array || oKind == reflect.Slice)) {
```
aschmahmann, line 160: This can be removed, but basically is just a hack for StringsOption so that it can still opaquely call this function and everything works. Doesn't seem too unreasonable though.

option.go
```diff
@@ -149,6 +149,18 @@ func NewOption(kind reflect.Kind, names ...string) Option {
 }
 
 func (o *option) WithDefault(v interface{}) Option {
+	if v == nil {
+		panic(fmt.Errorf("cannot use nil as a default"))
+	}
+
+	// if type of value does not match the option type
+	if vKind, oKind := reflect.TypeOf(v).Kind(), o.Type(); vKind != oKind {
```
aschmahmann, line 157: This feels like it alleviates a large footgun stemming from Go's lack of generics. Especially since in go-ipfs we build the commands at compile time it seems like panicking when someone has improperly set the default is just a good idea and will save us from mistakes we've made in the past such as using an `1` instead of a `uint(1)` default value in `UintOption` or using `"foo"` instead of `[]string{"foo"}` as a default value in StringsOption.

gammazero (approved):
LGTM.  Provided some feedback to your notes.

option.go
aschmahmann: @gammazero @lidel I pushed an update here that panics when an empty delimiter is passed, but still uses the same underlying struct for StringsOption and DelimitedStringsOption with StringsOption using an empty delimiter. let me know what you think and we'll merge 🚢

PR #205, opened 2021-02-09
WIP Modernize documentation and examples

djdv:
I keep referencing these docs which are several years out of date or missing. I'm going to start tracking my findings instead of forgetting them, but probably won't make a complete pass.
This is really rough right now but when more is done, I'll flip this from a draft.
(Not sure when I'll get to it but I'm going to be reviewing code that deals with this library soon)

In the meantime, anyone feel free to point things out here.
X is missing, Y is wrong, your update of Y is wrong, etc.

Commit history is going to be force pushed over randomly, so don't depend on it until we're out of draft at least.

- every docstring needs a pass, empty ones should be filled in
- I need to rewrite most of the new example, but it's a base

### DIFF
--- a/chan.go
+++ b/chan.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 )
 
+// TODO: no documentation; [54dbca2b-17f2-42a8-af93-c8d713866138]
 func NewChanResponsePair(req *Request) (ResponseEmitter, Response) {
 	r := &chanResponse{
 		req:     req,
--- a/command.go
+++ b/command.go
@@ -173,8 +173,8 @@ func (c *Command) call(req *Request, re ResponseEmitter, env Environment) error 
 	return cmd.Run(req, re, env)
 }
 
-// Resolve returns the subcommands at the given path
-// The returned set of subcommands starts with this command and therefore is always at least size 1
+// Resolve returns the subcommands at the given path.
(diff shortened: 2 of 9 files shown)


djdv:
Reminders:
- There should probably be an explicit remark around either [ReqLog](https://github.com/ipfs/go-ipfs-cmds/blob/4ade007405e5d3befb14184290576c63cc43a6a3/reqlog.go#L26-L27) and/or the [http.NewHandler](https://github.com/ipfs/go-ipfs-cmds/blob/4ade007405e5d3befb14184290576c63cc43a6a3/http/handler.go#L61-L62), which informs the user of [the relation between them](https://github.com/ipfs/go-ipfs-cmds/blob/4ade007405e5d3befb14184290576c63cc43a6a3/http/handler.go#L187).
  - also consider defining a public version of this interface rather than using an [adhoc one in `http`](https://github.com/ipfs/go-ipfs-cmds/blob/4ade007405e5d3befb14184290576c63cc43a6a3/http/handler.go#L85)
  - The opaqueness around `env` in the handler constructor should be made clear, the user should probably be informed that it's going to be passed to `Call`; as-is there's no remark, just pass in an empty interface{} and ???.

gammazero:
The updates here are now out of date. Also, this PR adds many TODOs instead of providing the documentation in the first place.

Closing in favor of new PR with updated documentation and without TODO's.

PR #209, opened 2021-05-02
Support encoing using IPLD-Prime codecs

willscott:
Allows for an argument of `--encoding=dag-json` or other valid ipld-prime codecs to be used in commands like `dag get`, where a dag node is being encoded to stdout.

Note: this relies on https://github.com/ipfs/go-ipfs/pull/7976 so that the node returned from the command is an ipld prime node that the encoder can successfully encode. When the return value is not an ipld prime node, and an ipld-prime codec is chosen, an error that an invalid node time has been attempted to be encoded is emitted.


encoding.go
```diff
@@ -172,7 +196,25 @@ func GetEncoder(req *Request, w io.Writer, def EncodingType) (encType EncodingTy
...
 	}
 	if !ok {
-		return encType, nil, Errorf(ErrClient, "invalid encoding: %s", encType)
+		// In addition to the explicitly named legacy encoders defined in this file,
+		// support encoding with the encoders registered in the ipld-prime multicodec
+		// registry.
+		ipldCodec, ok := mc.Of(string(encType))
+		if !ok {
+			if !ok {
+				n, err := strconv.Atoi(string(encType))
```
warpfork, line 205: This seems like two new features:

- more string names supported
- base-10 numbers supported

Are there docs changes that should go with it?

warpfork (approved):
As a novice reviewer to this repo -- this looks fine to me.

willscott:
we no longer need this PR based on the current state of https://github.com/ipfs/go-ipfs/pull/7995

PR #211, opened 2021-07-18
fix: CLI parsing of files should allow multiple files with the same name

aschmahmann:
fixes https://github.com/ipfs/go-ipfs/issues/8264

### DIFF
--- a/cli/parse.go
+++ b/cli/parse.go
@@ -332,7 +332,6 @@ func parseArgs(req *cmds.Request, root *cmds.Command, stdin *os.File) error {
 						return err
 					}
 
-					fpath = path.Base(fpath)
 					file = nf
 				}
 


Stebalien:
This is... going to be tricky. It produces an invalid "go-ipfs-files" directory and I'm _pretty_ sure it fails with `ipfs add -w`. Even with one file, that would produce a directory with a filename that potentially contains slashes (maybe...)

I wonder if renaming would be a better option?

aschmahmann:
> I wonder if renaming would be a better option?

Do you mean giving the files random names? That might work, but we'd have to check on the users of this to make sure we're not breaking them.

An alternative might be to use some structure other than a naive `files.NewMapDirectory` here to contain all the files.

Stebalien:
> Do you mean giving the files random names? That might work, but we'd have to check on the users of this to make sure we're not breaking them.

Well, rename with `.1` etc when we have a conflict.

But yeah, if we could just send the files as an array, effectively, that might work better.

BigLep:
Next step:  use some structure other than a naive files.NewMapDirectory here to contain all the files. For example, I'm not sure the HTTP API has this problem with it's multipart reader implementation.

aschmahmann:
> But yeah, if we could just send the files as an array

@Stebalien it seems like we have https://github.com/ipfs/go-ipfs-files/blob/fa4cddf7b3b8d69f6734bb6cc566620d5993407b/slicedirectory.go#L66. Should that be sufficient?

Issue #115, opened 2018-08-10
PostRun is an awkward fit for CLI output


Stebalien:
When writing CLI output, we often want to write to stdout/stderr, sometimes even controlling terminal output. For commands that *just* need to format to stdout, we can use a text marshaler. However, for commands that need to write to some parts to stderr and/or control the terminal, that doesn't really fit.

So, in practice, we use a `PostRun` function. However, that isn't a fit *either* as it expects us to transform a response emitter.

Currently, we just hack it by writing to stdout/stderr anyways but I'd like a better option.

---

So, I'd like to:

1. Deprecate PostRun. If you look at how we use it, we don't *actually* use it to transform one response emitter into another, we use it to write to the console. We could keep it, but, IMO, the current form doesn't fit very well. I can't think of a case where I'd want to transform the results based on the output mode.
2. Encourage using `Encoders` when we don't need anything fancy (just text and a terminal error).
3. Provide a DisplayCLI field that takes a `func(res cmds.Response, stdout, stderr io.Writer) error`.

In the future, we could add additional display functions for different interfaces (e.g., browsers) but this seems like a good starting point. Thoughts?

kevina:
Note that #102 changes the signature of PostRun so it does _not_ return a response emitter.

Stebalien:
No, but it *takes* a response and a response emitter. It still operates as a stream transform.

PR #215, opened 2021-08-19
CLI and API command output encoding

jbouwman:
Several open go-ipfs issues mention problems with interaction of the global `--encoding=` flag and the Encoders and PostRun fields of commands. This branch explores output encoding strategy across offline CLI, online CLI, and API modes using the approach suggested in #115.

Specific go-ipfs Issues:

- https://github.com/ipfs/go-ipfs/issues/7050 json encoding for `ls`
- https://github.com/ipfs/go-ipfs/issues/1121 json encoding for `add`
- https://github.com/ipfs/go-ipfs/issues/5594 json encoding for `stats bw`

Related Issues:

- https://github.com/ipfs/go-ipfs/issues/6640 global flags on subcommands

Related PRs:

- https://github.com/ipfs/go-ipfs/pull/5620 json for 'stat'


Stebalien (changes requested):
Thanks for implementing this, it's been bugging me for a while... And thanks for making it possible to incrementally switch over, that'll make this much easier to land.

Stebalien:
Sorry for the very slow response, things that touch this library tend to be scary so I tend to procrastinate. This looks quite reasonable, so I'll be sure to make future review cycles shorter.

jbouwman:
> Sorry for the very slow response, things that touch this library tend to be scary so I tend to procrastinate. This looks quite reasonable, so I'll be sure to make future review cycles shorter.

No worries whatsoever. Thanks very much for the review.

http/responseemitter.go
```diff
@@ -27,7 +27,7 @@ var (
 
 // NewResponseEmitter returns a new ResponseEmitter.
 func NewResponseEmitter(w http.ResponseWriter, method string, req *cmds.Request, opts ...ResponseEmitterOption) (ResponseEmitter, error) {
-	encType, enc, err := cmds.GetEncoder(req, w, cmds.JSON)
+	encType, enc, err := cmds.GetLocalEncoder(req, w, cmds.JSON)
```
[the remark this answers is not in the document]
Stebalien: > What do you think about running DisplayCLI in the case that an HTTP client requests text/plain

In that case, we should continue to use the text encoder. `DisplayCLI` is specifically for controlling a terminal when a simple text encoder isn't sufficient.

We _currently_ use a `PostRun` function (which is run entirely client-side) but that ends up breaking `ipfs ... --enc=json` (because we intercept the response from the server before it hits the local encoder).

executor.go
```diff
@@ -32,6 +35,33 @@ type executor struct {
 	root *Command
 }
 
+// GetLocalEncoder provides special treatment for text encoding
+// when Command.DisplayCLI field is non-nil, by defining an
+// Encoder that delegates to a nested emitter that consumes a Response
+// and writes to the underlying io.Writer using DisplayCLI.
+func GetLocalEncoder(req *Request, w io.Writer, def EncodingType) (EncodingType, Encoder, error) {
```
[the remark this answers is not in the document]
jbouwman: Terminal orientation makes sense, dropped this.

executor.go
```diff
@@ -88,7 +113,7 @@ func (x *executor) Execute(req *Request, re ResponseEmitter, env Environment) er
 	}
 
 	postRunCh := maybeStartPostRun(cmd.PostRun)
-	runCloseErr := re.CloseWithError(cmd.Run(req, re, env))
+	runCloseErr := re.CloseWithError(run(req, re, env))
 	postCloseErr := <-postRunCh
 	switch runCloseErr {
 	case ErrClosingClosedEmitter, nil:
```
Stebalien, line 119: We need to read from the display error channel, right?

djdv:
Sorry for the conflicts (；´∀｀)
Just as a heads up, feel free to keep or clobber whatever variable/function names were used in https://github.com/ipfs/go-ipfs-cmds/pull/216 and https://github.com/ipfs/go-ipfs-cmds/pull/217
The names don't matter compared to the fix itself, so the choice is yours when resolving the conflict.

Also, if this gets merged https://github.com/ipfs/go-ipfs-cmds/pull/219
it'll introduce another one. In `cli.Run` for this line `if encType == cmds.Text && !cmd.HasText() {`
If the linked PR is correct, you should be able to just drop the conflicting hunk in this PR when that happens.
(No guarantees that I'm right on that though, or that it gets merged in - but if so I see that you already handle this in `cli.Parse` too `if req.Command.HasText() {` )

I'll _try_ not to create any more trouble ;^)

lidel (commented):
Resolved conflicts. 

It seems https://github.com/ipfs/go-ipfs-cmds/pull/215#pullrequestreview-750133320 the only pending question / concern – @jbouwman mind taking a look at 
@Stebalien question?

jbouwman:
> Resolved conflicts. sparkles
> 
> It seems [#215 (comment)](https://github.com/ipfs/go-ipfs-cmds/pull/215#discussion_r705179052) the only remaining question / concern – @jbouwman mind taking a look at @Stebalien's question?

@lidel sure thing, done. @Stebalien does that look right, in terms of order of error checks?

executor.go
```diff
@@ -50,14 +51,45 @@ func (x *executor) Execute(req *Request, re ResponseEmitter, env Environment) er
 			return err
 		}
 	}
+
+	return EmitResponse(cmd.Run, req, re, env)
```
Stebalien, line 55: Sorry, I just realized we're doing this in the wrong place. `Execute` should just execute the command (pre, run, post) and feed the results to the response emitter.

I believe the best place to put it is in `cli.NewResponseEmitter`. That is, if the command defines a `DisplayCLI` function, `cli.NewResponseEmitter` would:

1. Call `NewChanResponsePair` to get an emitter/response.
2. Pass the response to `DisplayCLI`.
3. Return the emitter (you may need to wrap it in something? not sure).

That way, we only attempt to write to the terminal if we're actually running in a CLI.

Stebalien:
> does that look right, in terms of order of error checks?

Yes. But I missed the other thing.

jbouwman:
> > does that look right, in terms of order of error checks?
> 
> Yes. But I missed the other thing.

I had cycles to look at this again. The cli.ResponseEmitter wrapper for DisplayCLI seems a little odd -- its main role seems to be in allowing PostRun-enabled commands access to process exit value via a cast. I didn't see a less invasive way to do it though. I'd welcome your thoughts.

executor.go
jbouwman: Followed this approach.

BigLep:
@Stebalien : what are the recommended next steps?  Is this high enough priority for us to engage on right now?

BigLep:
@Stebalien : what are the recommended next steps?  Is this high enough priority for us to engage on right now?

BigLep:
@jbouwman : it looks like there are some comments from @Stebalien that you can incorporate (but I see there are open questions too).

@Stebalien : are you able to give definitive on what should be done here, or should we drop this for now due to not having bandwidth to engage?

BigLep:
@jbouwman : I have converted this back to draft.  Please mark it as "Ready for review" when comments have been incorporated.

Stebalien:
>  are you able to give definitive on what should be done here, or should we drop this for now due to not having bandwidth to engage?

I just don't have the bandwidth right now, unfortunately.

PR #217, opened 2021-09-08
Fix my own bug in postrun Execute

djdv:
I left in a bad condition in https://github.com/ipfs/go-ipfs-cmds/pull/216 which was supposed to be deleted.
This should amend that and adds tests to make sure postruns actually executes if it should.

>What was wrong?

The condition was supposed to be checking `formatters`, but was checking the wrong value.
>Why is this condition removed instead of changed?

We're reading from a map type, even if it's nil we'll just get back the 0 value when indexing it.
In this case that's a nil function pointer, which is checked before being assigned and later used.
`formatters[typer.Type()] != nil`

### DIFF
--- a/chan.go
+++ b/chan.go
@@ -96,6 +96,10 @@ func (r *chanResponse) Next() (interface{}, error) {
 		ctx = context.Background()
 	}
 
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
 	select {
 	case v, ok := <-r.ch:
 		if !ok {
@@ -144,6 +148,10 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 	}
 
 	ctx := re.req.Context
+
+	if err := ctx.Err(); err != nil {
+		return err
+	}
 
 	select {
 	case re.ch <- v:
--- a/command_test.go
+++ b/command_test.go
@@ -316,6 +316,7 @@ func TestCancel(t *testing.T) {
 	}
 
 	re, res := NewChanResponsePair(req)
+	cancel()
 
 	go func() {
 		err := re.Emit("abc")
@@ -327,8 +328,6 @@ func TestCancel(t *testing.T) {
 		re.Close()
 		close(wait)
 	}()
-
-	cancel()
 
 	_, err = res.Next()
 	if err != context.Canceled {
--- a/executor.go
+++ b/executor.go
@@ -56,25 +56,22 @@ func (x *executor) Execute(req *Request, re ResponseEmitter, env Environment) er
 			postRunCh = make(chan error)
 		)
 
+		// Check if we have a formatter for this emitter type.
+		typer, isTyper := re.(interface {
+			Type() PostRunType
+		})
+		if isTyper {
+			postRun = formatters[typer.Type()]
+		}
+		// If not, just return nil via closing.
 		if postRun == nil {
 			close(postRunCh)
 			return postRunCh
 		}
 
-		// check if we have a formatter for this emitter type
-		typer, isTyper := re.(interface {
-			Type() PostRunType
-		})
-		if isTyper &&
-			formatters[typer.Type()] != nil {
-			postRun = formatters[typer.Type()]
-		} else {
-			close(postRunCh)
-			return postRunCh
-		}
-
-		// redirect emitter to us
-		// and start waiting for emissions
+		// Otherwise, relay emitter responses
+		// from Run to PostRun, and
+		// from PostRun to the original emitter.
 		var (
 			postRes     Response
 			postEmitter = re
--- a/executor_test.go
+++ b/executor_test.go
@@ -114,3 +114,66 @@ func TestExecutorNotTyper(t *testing.T) {
 		t.Fatalf("expected EOF but got: %s", err)
 	}
 }
+
+func TestExecutorPostRun(t *testing.T) {
+	expectedValue := errors.New("postrun ran")
+	testCmd := &Command{
+		Run: func(*Request, ResponseEmitter, Environment) error {
+			return nil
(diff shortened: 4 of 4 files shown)


djdv:
@guseggert
Round 2. 😅

The testing code could probably be less copy-paste, but I think it at least works as expected.
I'm assuming it makes sense for `Execute` to block in this kind of setup, waiting for postrun to exit.
(and not blocking when nothing is trying to Emit)

Not sure if there's a better way than using a mocked `Typer` to test this. And maybe it shouldn't use the CLI type rather than something custom. But all these are little nits.
We can make any changes if they're desired.

djdv:
>A less roundabout way to do this might be to just return a defined error value from postRun and check for it.

I went ahead with this. It simplifies the test considerably and should be equally as valid. We're testing the method not the emitter anyway.
Diff: https://github.com/ipfs/go-ipfs-cmds/compare/bba17015f8a4668a0ce2dc63a508fd7a53c3bf41..8998fa443d492177589218f243ab3deba9b2f2cc

djdv:
For posterity I checked what [commit I was using in my own repo](https://github.com/djdv/go-ipfs-cmds/commit/92cf96be03f0) and why it worked where the one submitted in the PR didn't.
It's not really important other than for the context of it.

```diff
diff --git a/executor.go b/executor.go
index 77f58d5..5bcfb7a 100644
--- a/executor.go
+++ b/executor.go
@@ -50,24 +50,24 @@ func (x *executor) Execute(req *Request, re ResponseEmitter, env Environment) er
 			return err
 		}
 	}
-	maybeStartPostRun := func(postMap PostRunMap) <-chan error {
+	maybeStartPostRun := func(formatters PostRunMap) <-chan error {
 		var (
 			postRun   func(Response, ResponseEmitter) error
 			postRunCh = make(chan error)
 		)
 
-		if cmd.PostRun == nil {
+		if postRun == nil {
 			close(postRunCh)
 			return postRunCh
 		}
 
-		// check if we have a function for this emitter
+		// check if we have a formatter for this emitter type
 		typer, isTyper := re.(interface {
 			Type() PostRunType
 		})
 		if isTyper &&
-			cmd.PostRun[typer.Type()] != nil {
-			postRun = cmd.PostRun[typer.Type()]
+			formatters[typer.Type()] != nil {
+			postRun = formatters[typer.Type()]
 		} else {
 			close(postRunCh)
 			return postRunCh
```
This falls in line with my assumption; I meant to check `formatters` but referenced the wrong variable when I changed `maybeStartPostRun`'s signature.
Originally this was just a closure with no arguments, referencing `cmd` directly, so when I changed it to accept a `PostRunMap` as an argument, I should have dropped that check then instead of changing it. (like is done in this PR)

djdv (commented):
Good enough. 👌

djdv:
This is interesting.
https://github.com/djdv/go-ipfs-cmds/runs/3558361879
I'm pretty sure my patch is not causing this failure. Running locally this test passes most, but not all the time.
I looked into it and, I think `TestCancel` may be implemented wrong, depending on randomness in the scheduler.

`cancel` should likely be called before the `Emit` goroutine even has the potential to run.
But making this change still did not fix the problem.
Within `Emit`, and `Next` for the `chanResponseEmitter`, it's also depending on randomness within `select` to choose the context error over sending the value, which is not guaranteed in any way.
So I just check up front if we're canceled, and bail out then. While leaving the selects which block on communication.
So if the caller never reads the value, but does call cancel at that point, they'll still get a context error also.

`go test -v -race -run TestCancel -count 10000` would not pass for me before this on a Linux system, but now does.
So too does `go test -v -race -count 10000` for the whole package on Linux and Windows.

I added the commit here but we can separate it out if desired.

With the code being correct and CI passing, this should be ready for review for real now.

chan.go
```diff
@@ -145,6 +149,10 @@ func (re *chanResponseEmitter) Emit(v interface{}) error {
 
 	ctx := re.req.Context
 
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
```
djdv, line 155: The declaration of ctx, and this check should probably be hoisted to the top of the function.
If we're canceled, we probably shouldn't do anything other than return.
The locking, EmitChan, etc. seems pointless if we'll get here anyway.

executor.go
```diff
@@ -56,25 +56,22 @@ func (x *executor) Execute(req *Request, re ResponseEmitter, env Environment) er
...
+		if postRun == nil {
 			close(postRunCh)
 			return postRunCh
 		}
 
-		// redirect emitter to us
-		// and start waiting for emissions
+		// Otherwise, relay emitter responses
+		// from Run to PostRun, and
+		// from PostRun to the original emitter.
```
djdv, line 74: I think this makes sense, but that's because I wrote it.
If it could be reworded or is confusing, someone let me know.

command_test.go
```diff
@@ -316,6 +316,7 @@ func TestCancel(t *testing.T) {
 	}
 
 	re, res := NewChanResponsePair(req)
+	cancel()
```
djdv, line 319: >`cancel` should likely be called before the `Emit` goroutine even has the potential to run.

Incorrect. This needs to go back where it was.
We pass either way, but at least when `cancel` is called late, there's a 50/50 chance we also test that pending sends will be canceled instead of sent.
Depending on who runs first, `Emit` or `cancel`.

chan.go
guseggert: I'd rather just leave it next to the `select` like this, so that we preserve the existing post-cancellation side effects of consuming v if it's a chan, and unblocking `Length()`, which might be important somewhere else (don't know off the top of my head).

guseggert (approved):

PR #219, opened 2021-09-20
minor cleanup for cli.Run

djdv:
I was going over this section and I'm pretty sure all of it is redundant.
(Someone please double check it though)

For 6e8d6f4
The nil checks for `req` are removed because Parse states "This function never returns nil, even on error."
So these conditions should never change / need to be checked.

For c9f36de
`Parse` already does this internally (and in a more formal way) to the `Command` before returning it (attached as `req.Command`).
https://github.com/ipfs/go-ipfs-cmds/blob/87b5c50105f87fafa651000c4955571b132948b2/cli/parse.go#L48-L59


djdv:
Tiny force push for consistency.
`printMetaHelp` directly above `printHelp` was already referencing `.Path` directly.
This just matches that.
```diff
diff --git a/cli/run.go b/cli/run.go
index 567f3ac..0317cca 100644
--- a/cli/run.go
+++ b/cli/run.go
@@ -59,9 +59,7 @@ func Run(ctx context.Context, root *cmds.Command,
 			helpFunc = LongHelp
 		}
 
-		path := req.Path
-
-		if err := helpFunc(cmdline[0], root, path, w); err != nil {
+		if err := helpFunc(cmdline[0], root, req.Path, w); err != nil {
 			// This should not happen
 			panic(err)
 		}
```

djdv:
>For ~c9f36de~ 7dc3772
`Parse` already does this internally (and in a more formal way) to the `Command` before returning it (attached as `req.Command`).

Nope. This isn't correct.
While this does happen in `Parse`, it's only when `--enc` is not provided.

I'm going to look at this again later.
Either that commit can just be dropped, or a new one can replace it.
I'm thinking that `Parse` could probably be changed in a way that it handles this, but it might not make sense to.
Need to look at it with non-tired eyes.

djdv:
I'm pretty sure the change to `Parse` would be fine, but without good test coverage none of this is worth the risks.

PR #220, opened 2021-09-29
fix(cli): use NewSliceDirectory for duplicate file paths

schomatis:
fixes ipfs/go-ipfs#8264
alternative to https://github.com/ipfs/go-ipfs-cmds/pull/211

Being tested in https://github.com/ipfs/go-ipfs/pull/8493.

### DIFF
--- a/cli/parse.go
+++ b/cli/parse.go
@@ -9,6 +9,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"strings"
 
 	osh "github.com/Kubuxu/go-os-helper"
@@ -236,7 +237,8 @@ func parseArgs(req *cmds.Request, root *cmds.Command, stdin *os.File) error {
 	}
 
 	stringArgs := make([]string, 0, numInputs)
-	fileArgs := make(map[string]files.Node)
+	fileArgs := make([]files.DirEntry, 0)
+	var fileStdin files.Node
 
 	// the index of the current argument definition
 	iArgDef := 0
@@ -266,7 +268,7 @@ func parseArgs(req *cmds.Request, root *cmds.Command, stdin *os.File) error {
 				stringArgs, inputs = append(stringArgs, inputs[0]), inputs[1:]
 			} else if stdin != nil && argDef.SupportsStdin && !fillingVariadic {
 				if r, err := maybeWrapStdin(stdin, msgStdinInfo); err == nil {
-					fileArgs["stdin"], err = files.NewReaderPathFile(stdin.Name(), r, nil)
+					fileStdin, err = files.NewReaderPathFile(stdin.Name(), r, nil)
 					if err != nil {
 						return err
 					}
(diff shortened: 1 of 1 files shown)


cli/parse.go
```diff
@@ -332,7 +332,17 @@ func parseArgs(req *cmds.Request, root *cmds.Command, stdin *os.File) error {
 						return err
 					}
 
-					fpath = filepath.Base(fpath)
+					base := filepath.Base(fpath)
+					fpath = base
+					if _, ok := fileArgs[fpath]; ok {
+						// Ensure a unique fpath by suffixing ' (n)'.
```
aschmahmann, line 338: You can try a PR bubbling this into go-ipfs and adding some `ipfs add` tests to see what the outputs look like.

However, I'm pretty sure this will end up with confusing output names for users. For example adding `/foo/x`, `/bar/x`, and `/bar/x (1)` may (if I'm following the code correctly) end up showing outputs like:
```
/foo/x
/bar/x (1)
/bar/x (2)
```

which is weird.

Would just use a SliceDirectory (https://github.com/ipfs/go-ipfs-files/blob/544b04dda23a1637b6cb1966f78a68892177aff2/slicedirectory.go#L66) instead of a Map one work here? It seems like it's basically the intended use case.

Either way it seems like adding tests in go-ipfs is the next step to confirm end user behavior and if it's really what we want.

aschmahmann (changes requested):

cli/parse.go
schomatis: > which is weird.

Yes, but note this is already the existing behavior for URL resources with the same path name.

schomatis: > Either way it seems like adding tests in go-ipfs is the next step to confirm end user behavior and if it's really what we want.

@aschmahmann I'm not sure from the original issue what exactly is the expected output from adding files with the same name. Could you describe here the expected output from the issue's original example, please?
```
mkdir -p /tmp/x/{1,2}
echo x > /tmp/x/1/x
echo y > /tmp/x/2/x
ipfs add -n --raw-leaves -- /tmp/x/1/x /tmp/x/2/x
```

Something like:
```
added <CID> some text for output /tmp/x/1/x
added <CID> some other *different* text for output /tmp/x/2/x
```

aschmahmann: The format of the output is:
`added <CID> path-in-local-system`

For example 
```
ipfs add -n --raw-leaves -- /tmp/x/1/x /tmp/x/2/x
added bafkreidtzm4frjuhvbeuzizsgbjqcyuc6pnnhhkcz5rmuttz3wrkvr6zvq /tmp/x/1/x
added bafkreib3wkv3nhv3e7574y6hmolcjrxmlyzrxba2lpemh26bbojil2iio4 /tmp/x/2/x
```

The issue is that currently just because the final path component (i.e. the file name) is the same in both cases we just drop one of them instead of adding both files

schomatis: Thanks, will look at the SliceDirectory proposed alternative then.

BigLep:
2021-11-12 conversation: needs review

BigLep:
@schomatis : how does this issue relate to https://github.com/ipfs/go-ipfs/pull/8493 ?

schomatis:
@BigLep Imported by it. The go-ipfs PR is basically a wrapper pointing to the changes in the dependency here where the actual logic was implemented.

guseggert (approved):

PR #223, opened 2022-01-18
feat: add command output option

schomatis:
Add an option to save command output to a file instead of the `stdout` argument (which is always set to stdout by `go-ipfs` consumer).

### DIFF
--- a/opts.go
+++ b/opts.go
@@ -12,10 +12,13 @@ const (
 	OptLongHelp  = "help"
 	DerefLong    = "dereference-args"
 	StdinName    = "stdin-name"
-	Hidden       = "hidden"
-	HiddenShort  = "H"
-	Ignore       = "ignore"
-	IgnoreRules  = "ignore-rules-path"
+	// FIXME(BLOCKING): Review names. (We may be able to use "output" once
+	//  it's cleaned from subcommands.)
+	OutputFile  = "command-output"
+	Hidden      = "hidden"
+	HiddenShort = "H"
+	Ignore      = "ignore"
+	IgnoreRules = "ignore-rules-path"
 )
 
 // options that are used by this package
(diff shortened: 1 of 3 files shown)


lidel (changes requested):
@aschmahmann you may have better intuition about what would be the most useful on windows.

I suspect stdout is not enough. My concern is that stderr would not end up in the (empty) file, but would go to terminal:

```console
$ ipfs id --command-output=stdout.txt
Error: no IPFS repo found in /home/lidel/tmp/go-ipfs-out.
please run: 'ipfs init'

$ ipfs get invalidcid --command-output=stdout.txt
Error: invalid path "invalidcid": selected encoding not supported
```

If someone runs a command in bash script, they may not monitor stderr and the message may be lost.

Perhaps an elegant way of handling this is to have explicit  `--stdout`  (or `--output`) and `--stderr` (or `--errors`) ?

```console
$ ipfs get invalidcid --stdout=stdout.txt --stderr=stderr.txt
$ cat stdout.txt 
$ cat stderr.txt
Error: invalid path "invalidcid": selected encoding not supported
```

schomatis:
@lidel Added error output option (we can review the option names at the end; I'm leaving for now the conspicuous `--command-output/error` for visibility).

opts.go
```diff
@@ -12,10 +12,14 @@ const (
...
 	DerefLong    = "dereference-args"
 	StdinName    = "stdin-name"
-	Hidden       = "hidden"
-	HiddenShort  = "H"
-	Ignore       = "ignore"
-	IgnoreRules  = "ignore-rules-path"
+	// FIXME(BLOCKING): Review names. (We may be able to use "output" once
+	//  it's cleaned from subcommands.)
+	OutputFile  = "command-output"
+	ErrorFile   = "command-error"
```
lidel, line 18: Let's keep this simple and use  `--stdout` and `--stderr`  
(self-explanatory + won't clash with `output` or other popular words)

```suggestion
	StdoutFile  = "stdout"
	StderrFile  = "stderr"
```

lidel (changes requested):

Issue #93, opened 2018-03-29
Deprecated flags/options


Stebalien:
It would be nice to support both deprecated and removed flags. Deprecated flags would be hidden (but still work) and "removed" flags would inform the user that the flag has been removed and why.

lidel:
Need to bump this – adding to maintenance board.


### Context

We have a bunch  of deprecated command families (`object`, `tar`) which are hidden from `ipfs --help`, but are still accessible + people see them as regular commands in docs which are generated from sources (https://docs.ipfs.io/reference/http/api/, created by generator in https://github.com/ipfs/ipfs-docs/tree/main/tools/http-api-docs). 

### Implementation  

We should add the ability to programmatically indicate three types of commands: 
- deprecated
- removed
- experimental (for new stuff, that may get changed)

Having that, `--help` and generated docs would be augmented to indicate that:
- `Helptext`  would be prefixed with `WARNING: EXPERIMENTAL` or `DEPRECATED`
- docs would have separate sections at the end for those types of commands, setting proper devexp expectations.

It could be as simple as adding three bool fields to `Command` struct, and wiring things up around that.

@schomatis would you be interested in adding this to your queue? 
Not a high priority, but would make a big difference for users learning-by-reading-docs or --help.

lidel:
This is implemented in:
- https://github.com/ipfs/go-ipfs-cmds/pull/225

And then leveraged in:
- https://github.com/ipfs/go-ipfs/pull/8802
- https://github.com/ipfs/ipfs-docs/pull/1086

PR #225, opened 2022-02-01
feat(Command): add status for the helptext

schomatis:
The logic of this patch is trivial; this review is more about adjusting to the desired presentation (if there are aesthetic changes needed an example of the desired output would be useful).

Closes https://github.com/ipfs/go-ipfs-cmds/issues/93.

### DIFF
--- a/cli/helptext.go
+++ b/cli/helptext.go
@@ -28,16 +28,20 @@ const (
 )
 
 type helpFields struct {
-	Indent      string
-	Usage       string
-	Path        string
-	Tagline     string
-	Arguments   string
-	Options     string
-	Synopsis    string
-	Subcommands string
-	Description string
-	MoreHelp    bool
+	Indent                  string
+	Warning                 string
+	Usage                   string
+	Path                    string
+	Tagline                 string
+	Arguments               string
+	Options                 string
+	Synopsis                string
+	Subcommands             string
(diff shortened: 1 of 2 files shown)


cli/helptext.go
```diff
@@ -445,6 +493,23 @@ func subcommandText(width int, cmd *cmds.Command, rootName string, path []string
...
+func generateWarningText(cmd *cmds.Command) string {
+	switch cmd.Status {
+	case cmds.Active:
+		return "" // We don't print a warning for a normal active command.
+	case cmds.Deprecated:
+		return "DEPRECATED, command will be removed in the future"
+	case cmds.Experimental:
+		return "EXPERIMENTAL, command may change in future releases"
+	case cmds.Removed:
+		return "REMOVED, command is no longer available"
```
lidel, line 507: IF this text looks ok, then we can merge and ship this.

lidel (approved):
Thank you @schomatis,  I took it for a spin in https://github.com/ipfs/go-ipfs/pull/8802 and https://github.com/ipfs/ipfs-docs/pull/1086, and it works perfectly :ok_hand:

cli/helptext.go
schomatis: Yep, looks fair.

PR #222, opened 2021-12-16
allow header and set header in client

hunjixin:
need send Authorization header to remote exec cmds. 
allow header and support set header value in client

PR #226, opened 2022-02-12
allow header and set header in client

Stebalien:
need send Authorization header to remote exec cmds.
allow header and support set header value in client

#222 with an extra commit.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -41,8 +41,8 @@ func ClientWithUserAgent(ua string) ClientOpt {
 	}
 }
 
-// ClientWithHeaders specifies the HTTP header for the client.
-func ClientWithHeaders(key, value string) ClientOpt {
+// ClientWithHeader adds an HTTP header to the client.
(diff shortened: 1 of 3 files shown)


mar1n3r0:
We could add the default allowed headers to [[reforigin_test](https://github.com/ipfs/go-ipfs-cmds/blob/master/http/reforigin_test.go)](https://github.com/ipfs/go-ipfs-cmds/blob/40b8fddb899fea1e59ca91e2d97bbea0c1e88689/http/reforigin_test.go#L53)

```
func originCfg(origins []string) *ServerConfig {
	cfg := NewServerConfig()
	cfg.SetAllowedOrigins(origins...)
	cfg.SetAllowedMethods("GET", "PUT", "POST")
        cfg.AddAllowedHeaders("Authorization", "Content-Type", "Content-Disposition") // etc any other sane defaults
	cfg.AllowGet = true
	return cfg
}
```

PR #230, opened 2022-04-05
fix(cli/parse): extract dir before name

schomatis:
Syncing to commit https://github.com/protocol/.github/commit/2aab7c039d804a6fcc9d4b330ef23e7788b4291f.

### DIFF
--- a/cli/parse.go
+++ b/cli/parse.go
@@ -329,8 +329,8 @@ func parseArgs(req *cmds.Request, root *cmds.Command, stdin *os.File) error {
 						return err
 					}
 
-					fpath = filepath.Base(fpath)
 					importDir := filepath.Dir(fpath)
+					fpath = filepath.Base(fpath)
 					prevDir, ok := fileImportDirName[fpath]
 					if !ok {
 						fileImportDirName[fpath] = importDir
--- a/version.json
+++ b/version.json
@@ -1,3 +1,3 @@
 {
-  "version": "v0.8.0"
+  "version": "v0.8.1"
 }


Jorropo (dismissed):
The change looks good, however this error message is terribely spammy:
```
Error: file name x repeated under different import directories: dir2 and dir1

USAGE
  ipfs add <path>... - Add a file or directory to IPFS.

  ipfs add [--recursive | -r] [--dereference-args] [--stdin-name=<stdin-name>] [--hidden | -H] [--ignore=<ignore>]... [--ignore-rules-path=<ignore-rules-path>] [--quiet | -q] [--quieter | -Q] [--silent]
           [--progress | -p] [--trickle | -t] [--only-hash | -n] [--wrap-with-directory | -w] [--chunker=<chunker> | -s] [--pin=false] [--raw-leaves] [--nocopy] [--fscache] [--cid-version=<cid-version>]
           [--hash=<hash>] [--inline] [--inline-limit=<inline-limit>] [--] <path>...

  Adds the content of <path> to IPFS. Use -r to add directories (recursively).

  'ipfs add <subcmd> --help'


```
Dumping the help isn't helpfull and dilute the signal.
I would expect the same thing as other APIs:
```console
$ ./ipfs pin add Qmfoo
Error: invalid path "Qmfoo": selected encoding not supported
```

schomatis:
@Jorropo I don't understand exactly what's your requested change here. The above is the behavior for _any_ command error, e.g., `ipfs add foo`. You should submit an independent issue to discuss that.

Jorropo (approved):
LGTM then

PR #231, opened 2022-05-20
feat(Command): filter subcommands

schomatis:
For https://github.com/ipfs/go-ipfs/issues/8059#issuecomment-1122437673.


schomatis:
@lidel I'm fine bumping this all the way to `Suggested version: v0.9.0`, your call.

PR #243, opened 2023-08-16
fix!: client with raw abs path option

hacdias:
See https://github.com/ipfs/kubo/pull/10068.

- [x] https://github.com/ipfs/boxo/pull/434
- [x] ⚠️ **Update to Boxo release before merge**\

Will be released in subsequent PR.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -132,7 +132,7 @@ func (re *writerResponseEmitter) SetLength(length uint64) {
 		return
 	}
 
-	*re.length = length
+	re.length = &length
 }
 
 func (re *writerResponseEmitter) Close() error {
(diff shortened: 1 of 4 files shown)


writer.go
```diff
@@ -132,7 +132,7 @@ func (re *writerResponseEmitter) SetLength(length uint64) {
 		return
 	}
 
-	*re.length = length
+	re.length = &length
```
Jorropo, line 135: Can you split this in it's own commit in case this SOMEHOW was correct and we want to revert your change ?

hacdias: (It was literally panicking)

hacdias: (But sure)

Issue #249, opened 2024-04-05
Http handler returns 200 OK on command panic


fabiobozzo:
My team's fork of Kubo runs on a few custom interface implementations, notably the `Pinner`. 
Recently, we found a bug in one of them which caused a panic, in the context of a command.

We were surprised to find out that the http API returns 200 OK, in case of panic.  
In order to demonstrate that behaviour, I opened a PR, with a couple of new http handler tests. 
See: https://github.com/ipfs/go-ipfs-cmds/pull/250

lidel:
Thank you, the remaining work will land in https://github.com/ipfs/kubo/issues/10398

PR #250, opened 2024-04-05
fix(http): return error in case of panic

fabiobozzo:
Address issue: https://github.com/ipfs/go-ipfs-cmds/issues/249

The new tests cover a case of sync http commands (e.g. `/cat`) and 'progress' commands (e.g. `/add`) panicking.


Stebalien (approved):

Stebalien:
Thanks!

Stebalien:
(and thanks for writing a test)

PR #252, opened 2024-04-15
chore: update deps

Stebalien:


Stebalien:
> I think this should be fine, unless there's some obscure Kubo test that expects 200 instead of 204.

There probably is. But this needs to be a minor version anwyays...

PR #254, opened 2024-05-06
Preparing a `v2.0.2+incompatible` pre-release to properly retract `v2.0.1+incompatible`

AnomalRoil:
We need to issue a higher version tag which self-retract itself in order to retract `v2.0.1+incompatible`
I have chosen a pre-release `v2.0.2+incompatible` value so that it will still be possible to create v2.0.x versions in the future without conflicting with it.

After merging this, please tag the latest master with tag `v2.0.2+incompatible` and push the tag to github 🙏🏻 

As per the Go spec:
> To retract a version, a module author should add a retract directive to go.mod, then publish a new version containing that directive. The new version must be higher than other release or pre-release versions; that is, the `@latest` [version query](https://go.dev/ref/mod#version-queries) should resolve to the new version before retractions are considered
> A version containing retractions may retract itself. If the highest release or pre-release version of a module retracts itself, the `@latest` query resolves to a lower version after retracted versions are excluded.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,4 +25,4 @@ retract v1.0.22 // old gx tag accidentally pushed as go tag
 
 retract v2.0.1+incompatible // old gx tag
 
-retract v2.0.2-retract // we need to use a newer version than v2.0.1 to retract v2.0.1+incompatible, but we can retract ourself directly once done
+retract v2.0.2+incompatible // we need to use a newer version than v2.0.1 to retract v2.0.1+incompatible, but we can retract ourself directly once done
--- a/go.mod
+++ b/go.mod
@@ -24,3 +24,5 @@ require (
 retract v1.0.22 // old gx tag accidentally pushed as go tag
 
 retract v2.0.1+incompatible // old gx tag
+
+retract v2.0.2-retract // we need to use a newer version than v2.0.1 to retract v2.0.1+incompatible, but we can retract ourself directly once done


go.mod
```diff
@@ -24,3 +24,5 @@ require (
 retract v1.0.22 // old gx tag accidentally pushed as go tag
 
 retract v2.0.1+incompatible // old gx tag
+
+retract v2.0.2-retract // we need to use a newer version than v2.0.1 to retract v2.0.1+incompatible, but we can retract ourself directly once done
```
aschmahmann, line 28: CI is failing with: `go.mod:28: retract github.com/ipfs/go-ipfs-cmds: version "v2.0.2-retract" invalid: should be v0 or v1, not v2`

It looks like this might be because we'd need to change the import path in this go mod file as well https://github.com/golang/go/issues/35732#issuecomment-584319096. I haven't dealt with two many v2 modules like this before though so lmk if I'm missing something.

aschmahmann: Side note: Does this need to merge into master or can this merge into a `v2.0.2-retract-branch` and be released from there?

aschmahmann (changes requested):

lidel:
Triage notes:

- @AnomalRoil where does this problem manifest? 
- gx tags were created before modern v2 conventions were introduced and it may be tricky to fix, so we want to understand the priority here

AnomalRoil:
Currently the Go ecosystems recognizes the v2.0.1-incompatible to be the latest release:
image

To properly retract a release you need to tag a higher version number tag with a retract statement for both.

See doc here: https://go.dev/ref/mod#go-mod-file-retract

lidel (approved):
Triage notes:
- Using https://go.dev/ref/mod#incompatible-versions should make this relatively safe in module version ecosystem, we will give it a try.

lidel:
Just for the record (cc @gammazero) this did not work because `+incompatible` is respected only if tag does not include `go.mod`:

```console 
$ GOPROXY=https://proxy.golang.org GO111MODULE=on go get "github.com/ipfs/go-ipfs-cmds@v2.0.2"
go: github.com/ipfs/go-ipfs-cmds@v2.0.2: reading https://proxy.golang.org/github.com/ipfs/go-ipfs-cmds/@v/v2.0.2.info: 404 Not Found
	server response: not found: github.com/ipfs/go-ipfs-cmds@v2.0.2: invalid version: module contains a go.mod file, so module path must match major version ("github.com/ipfs/go-ipfs-cmds/v2")
```

I've removed the tag since it is useless anyway and filled https://github.com/golang/go/issues/69093 to remove broken version from x/pkgsite.

AnomalRoil:
@lidel  I suspect this did not work properly because you tagged v1.13 instead of tagging v2.0.2+incompatible and v1.13 < v2.0.2+incompatible as per semver.

PR #276, opened 2025-01-07
Support HTTPS when using a specific API instance

acejam:
The `--api` flag in Kubo is not currently usable with an API instance behind TLS. This change allows HTTPS to be used.


lidel:
Thank you for opening this. 

I am afraid hardcoding port is not the right approach – we should simply respect explicitly passed  `https://` scheme, when passed, and make Kubo pass it, when addr passed to `ipfs --api` has `/https` or `/tls/http`. 

Let's continue in https://github.com/ipfs/go-ipfs-cmds/pull/277

PR #282, opened 2025-02-22
feat(errors): add dedicated error type for blocked content

PsychoPunkSage:
Changes:
- Add new `ErrBlocked` error type in `errors.go` 
- Handle HTTP 410 (Gone) and 451 (Unavailable For Legal Reasons) status codes in response parsing
- Map blocked content errors to appropriate status codes in response emitter

This provides proper error types for content blocking rather than relying on string matching. Works in conjunction with error type improvements in boxo.

Inspired by:  ipfs/boxo#591
Related to: ipfs/boxo#855

### DIFF
--- a/error.go
+++ b/error.go
@@ -22,6 +22,7 @@ const (
 	// ErrForbidden is returned when the client doesn't have permission to
 	// perform the requested operation.
 	ErrForbidden
+	ErrBlocked
 )
 
 func (e ErrorType) Error() string {
@@ -40,6 +41,8 @@ func (e ErrorType) String() string {
 		return "rate limited"
 	case ErrForbidden:
 		return "request forbidden"
+	case ErrBlocked:
+		return "content blocked"
 	default:
 		return "unknown error code"
 	}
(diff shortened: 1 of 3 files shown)


lidel (changes requested):
Thank you. We need to see how this will work in practice.
Namely, who throws `ErrBlocked`?

Similar request to https://github.com/ipfs/boxo/pull/855#pullrequestreview-2646177585 – mind opening a companion PR against https://github.com/ipfs/kubo that uses go-ipfs-cmds from this PR and demonstrates how this should be used?

We have RPC/CLI tests in https://github.com/ipfs/kubo/blob/baa94fcb2f8ff0e23a9da8e7934c05ad5540afe5/test/cli/content_blocking_test.go#L186 that expect specific string.

What needs to be added is additional HTTP RPC test there that makes request to `/api/v0/{cmd}` and then inspects HTTP response code.

Once we have it in Kubo, we can enable it in this repository in [.github/workflows/kubo-sharness.yml](https://github.com/ipfs/go-ipfs-cmds/blob/master/.github/workflows/kubo-sharness.yml).

gammazero:
Closing to prefer a more unified error strategy.

PR #285, opened 2025-05-12
Update go log v2

gammazero:


gammazero:
@lidel The only issue here is the removal of the call to `ContextWithLoggable`, [here](https://github.com/ipfs/go-ipfs-cmds/pull/285/files#diff-be4b2ce0063f079fc9f70549add23647bacd60efef91f0684e0cff32ee4efb5eL168)

I think this is save enough to remove, but where should its removal be documented?

guillaumemichel (approved):

PR #303, opened 2025-09-29
Propagate trace information

dennis-tra:
This PR adds trace propagation capabilities to the default HTTP client.

The `otelhttp.DefaultClient` defines a custom `http.RoundTripper` that extracts trace information from the context and populates the HTTP request header with `traceparent` and `tracestate` data which is then passed to kubo. Kubo is already using the [otel HTTP handler](https://github.com/ipfs/kubo/blob/13fbb76de4dd0ba89b573371f7b3cf8a9c0b07fd/core/corehttp/commands.go#L149), so there the data is extracted from the header and passed into the request context.

That way users of this go-ipfs-cmds library can more easily correlate issued commands with kubo actions.

Passing a custom http client via [ClientWithHTTPClient](https://github.com/probe-lab/go-ipfs-cmds/blob/2d2e31b85ee48874f0097c8938c604c4e8c88b01/http/client.go#L57) will overwrite this configuration.


dennis-tra:
Now that I think about it, should the user just pass in a custom Client instead? Maybe that's the better option

